PR #9, opened 2020-02-26
add reference to boolean in function GetIsFieldNumeric

chenhend:
This PR resolved the following task: 1667

### DIFF
--- a/K2Bridge/Visitors/QueryStringClauseVisitor.cs
+++ b/K2Bridge/Visitors/QueryStringClauseVisitor.cs
@@ -104,6 +104,7 @@ namespace K2Bridge.Visitors
                 "long" => true,
                 "float" => true,
                 "double" => true,
+                "boolean" => true,
                 _ => false,
             };
         }
--- a/K2Bridge/Visitors/QueryStringClauseVisitor.cs
+++ b/K2Bridge/Visitors/QueryStringClauseVisitor.cs
@@ -44,10 +44,17 @@ namespace K2Bridge.Visitors
             switch (queryStringClause.ParsedType)
             {
                 case QueryStringClause.Subtype.Term:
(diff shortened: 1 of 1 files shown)


K2Bridge/Visitors/QueryStringClauseVisitor.cs
```diff
@@ -44,10 +44,17 @@ public void Visit(QueryStringClause queryStringClause)
             switch (queryStringClause.ParsedType)
             {
                 case QueryStringClause.Subtype.Term:
-                    var isNumeric = GetIsFieldNumeric(queryStringClause.ParsedFieldName).Result;
-                    if (isNumeric)
+                    var isMatchPhraseFieldType = IsMatchPhraseFieldType(queryStringClause.ParsedFieldName).Result;
```
eladiw: This area will conflict with PR: https://github.com/microsoft/K2Bridge/pull/12
I am marking with do not merge so we will not enter this before.

K2Bridge/Visitors/QueryStringClauseVisitor.cs
```diff
@@ -44,10 +44,17 @@ public void Visit(QueryStringClause queryStringClause)
...
+                    if (isMatchPhraseFieldType)
                     {
-                        queryStringClause.KustoQL = $"{queryStringClause.ParsedFieldName} {KustoQLOperators.Equal} {queryStringClause.Phrase}";
+                        var matchPhraseClause = new MatchPhraseClause()
+                        {
+                            FieldName = queryStringClause.ParsedFieldName,
+                            Phrase = queryStringClause.Phrase,
+                        };
+
+                        matchPhraseClause.Accept(this);
```
eladiw: on another PR (not yet open) we are adding support for >= and <=

what happens if the field type is numeric but the pharse is '>=2', this is a current bug.
so you will get to MatchPharseClause and you will need to support >= in addition to Equal.

PR #21, opened 2020-03-02
Update documentation and install commands

tamirkamara:
* Update install commands and Helm notes


docs/installation.md
```diff
@@ -49,20 +50,21 @@ If you need to build the image, please follow the [build instructions](./build.m
         ADX_TENANT_ID=[SERVICE_PRINCIPAL_TENANT_ID]
         ```
 
-        Optional - Enable ApplicationInsights telemetry
+        Optional - enable ApplicationInsights telemetry
+
         ```sh
         APPLICATION_INSIGHTS_KEY=[INSTRUMENTATION_KEY]
         COLLECT_TELEMETRY=true
```
boazsha: Should we add `enableQueryLogging`?

tamirkamara: It's optional per the recommendation to have it as an opt-in feature

guregini: But let's add it as an optional argument with '[ ]' to the command "helm install k2bridge..." : 
[--set settings.enableQueryLogging="true"]

tamirkamara: ok, so I'll remove the "optional section" since it'd be in the command

tamirkamara: Done. Please approve.

tamirkamara: I think I miss understood. enableQueryLogging is true by default, we don't want to encourage disabling it but we do document the option in the extended table linked from the main installation readme.
What do you think?

guregini: OK, so no change is needed (given enableQueryLogging is true by default)

PR #25, opened 2020-03-14
Streamline build stages

algattik:

* Build all docker images in a single job, to avoid duplicating build steps (layer reuse) and speed
up build.
* Parallelize Cleanup and Publish artifacts in concurrent stages to speed up completion.
* Merged build and security analysis stages into a single stage, to simplify the build graph.
* Removed obsolete references to non-existing variable GenerateNS.KUBERNETES_NAMESPACE

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -218,7 +218,6 @@ stages:
   - job: Deploy_Kusto
     displayName: Provision a new Kusto database
     condition: or(and(succeeded(), not(eq(variables['KUSTO_DB'], 'demo'))), and(succeeded(), eq(variables['KUSTO_DB'], 'demo'), variables['RUN_CREATE_DEMO_KUSTO']))
-    dependsOn: Terraform
     steps:
 
         - task: AzureCLI@1
@@ -232,7 +231,6 @@ stages:
 
   - job: Deploy_K2Bridge
     displayName: Deploy K2Bridge
-    dependsOn: Terraform
     variables:
       ${{ if ne(variables['Build.SourceBranchName'], 'master') }}:
         NOT_MASTER_ES_CONFIG: "--set elasticsearch.replicas=2 --set elasticsearch.minimumMasterNodes=2"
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -159,50 +150,35 @@ stages:
...
-        scriptLocation: inlineScript
-        inlineScript: |
-          # Push Helm chart
+          # Push Helm chart to ACR
           helm repo add elastic https://helm.elastic.co
           helm repo update
           helm dependency update charts/k2bridge
           empty_dir="$(Build.StagingDirectory)/charts_out"
           helm package --version "$SEMANTIC_VERSION" charts/k2bridge -d $empty_dir
           az acr helm push --force "$(ls $empty_dir/*)"
```
tamirkamara: We like to keep each function in its own task to make it easier to understand what failed before digging into the logs.
Can you bring those back?

azure-pipelines.yml
```diff
@@ -159,50 +150,35 @@ stages:
...
-  dependsOn:
-  - security_analysis
-  - build
+- stage: terraform
+  displayName: Terraform
+  dependsOn: []
   jobs:
-
   - job: Terraform
-    displayName: Prepare Terraform
```
tamirkamara: Can you give the job a name like "Prepare Tests Env"?

azure-pipelines.yml
```diff
@@ -159,50 +150,35 @@ stages:
...
+          # Build end-to-end test image
+          docker build --target end2endtest \
+            --build-arg VersionPrefix="$SEMANTIC_VERSION" \
+            -t "$ACR_NAME.azurecr.io/k2bridge-test:$SEMANTIC_VERSION" \
+            .
 
-          # Push Docker image
+          # Push Docker image to ACR
           docker push "$ACR_NAME.azurecr.io/k2bridge:$SEMANTIC_VERSION"
+          docker push "$ACR_NAME.azurecr.io/k2bridge-test:$SEMANTIC_VERSION"
```
tamirkamara: This was originally removed from the "build" stage since it takes more time and delays the overall progress of this pipeline. Why have you returned this?

azure-pipelines.yml
```diff
@@ -332,47 +309,12 @@ stages:
...
-      displayName: Build E2E-Test Image
-
-    - task: AzureCLI@1
-      displayName: Push Image
-      inputs:
-        azureSubscription: $(ACR_PUSH_SERVICE_CONNECTION)
-        scriptLocation: inlineScript
-        inlineScript: |
-          docker push "$ACR_NAME.azurecr.io/k2bridge-test:$SEMANTIC_VERSION"
-
```
tamirkamara: See my comment above in the build stage.
How about keeping this here to save time?

PR #40, opened 2020-04-21
Add a flag to disable terraform task

tamirkamara:

* Add a variable to the pipeline so the Terraform job as well as CredScan one could be disabled since they are currently not available on the msazure DevOps organization.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,7 +19,7 @@ stages:
 
   - job: security_analysis
     displayName: Security Analysis
-
+    condition: not(variables['SKIP_MSAZURE'])
     pool:
       # CredScan only runs on Windows
       vmImage: 'windows-latest'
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -19,7 +19,7 @@ stages:
 
   - job: security_analysis
     displayName: Security Analysis
-
+    condition: not(variables['SKIP_MSAZURE'])
```
eladiw, line 22: why not skip_terraform?

tamirkamara: there're 2 think to skip - terraform and credscan. I can separate to 2 variables but I thought that it would be simpler this way.

boazsha: I agree a single variable is better, also not sure about the name.
Maybe: `SKIP_UNSUPPORTED_PIPELINE_COMPONENTS`?

Issue #26, opened 2020-03-16
Doublequotes escaping is broken


spacentropy:
Hi. It seems that double-quotes escape in filters doesnt'work:
Repro:
Get field and value with double-quotes, click "+", receive error.
<img width="963" alt="Untitled" src="https://user-images.githubusercontent.com/464801/76792241-ac052480-67d3-11ea-87cf-8064ac76e8dc.png">
The error itself:

```
Error: Request to Elasticsearch failed: {"error":{"root_cause":[{"type":"SyntaxException","reason":"Syntax error: Query could not be parsed: I could not parse that, sorry.. Query: 'let fromUnixTimeMilli = (t:long) {datetime(1970 - 01 - 01) + t * 1millisec};\nlet _data = database(\"database\").Table | where (message == \"Health check \"LF\" completed after 35.422ms with status Unhealthy and '\"{\\\"Url\\\":\\\"http://url\\\"}\"'\") and (Timestamp >= fromUnixTimeMilli(1584385027085) and Timestamp <= fromUnixTimeMilli(1584385927085));\n(_data | summarize count() by Timestamp = bin(Timestamp, 30s)\n| order by Timestamp asc | as aggs);\n(_data | order by Timestamp desc | limit 500 | as hits)'","index_uuid":"index","index":"index"}],"type":"QueryException","reason":"Failed executing azure data explorer query","phase":"query"},"status":500}
```

spacentropy:
Here's repro:

```kql
.create table Purchases ( str:string, x:int)

.ingest inline into table Purchases <|
Shoes,1000
Wide Shoes,50
"Coats, black",20
"Coats with ""quotes""",5
```

Add this table to kibana, then click on `+` on `Coats with "quotes"` row
<img width="864" alt="Untitled" src="https://user-images.githubusercontent.com/464801/76889133-1bdbe380-6896-11ea-8765-6e8334d7789c.png">

AsafMah:
Fixed with #156

PR #41, opened 2020-05-29
Add dynamic columns filtering, fix doublequotes escaping.

### LINKED ISSUE #2: Add support for dynamic columns
Dynamic columns, meaning json objects stored in ADX are partially supported.
One can view and search them by referencing properties directly in the search box, but cannot use UI features such as filtering

spacentropy:
Fixes #2 and #26.
This solves the problem of filtering the nested JSON/dynamic fields for us.

**The logic behind this**
While getting FieldCapabilities if the table or function has a column with type "System.Object", we recursively perform `getschema` on 8 sampled rows of these columns(including nested ones). We add every possible `dynamic` property to the FieldCapabilities response, nesting them under main columns. The samples are bounded by 4h of lookback time, property depth is limited at 4.
I.e. we have such document in a column named `JSON`:

```json
{
	"Payload": {
		"eventId": {
			"Id": 108,
			"Name": "GeneratedHeaderUsingProvider"
		}
	}
}
```

When you try to get FieldCapabilities on index with this document in dynamic field, the code will look into 8 samples of this column JSON, `bag_unpack | getschema` will return that it has a sub-property `Payload`. Then it will recursively look into `JSON.Payload.eventId`, which in turn will have two columns with types int and string.

Kusto query that finds the properties looks like this:

```KQL
JSON 
| where ingestion_time() > ago(4h) 
| where isnotempty(Payload.eventId) 
| sample 16 
| project K2BridgeGetSchemaColumn=Payload.eventId  
| evaluate bag_unpack(K2BridgeGetSchemaColumn) 
| getschema 
| project ColumnName = strcat('Payload.eventId','.',ColumnName),ColumnType = DataType
```

We would get these fields added as FieldCaps to Kibana index mapping and the filter will work:

```
JSON.Payload.eventId.Id int
JSON.Payload.EventId.Name string
```

**Why `try-catch` while getting fields from Kusto?** 
https://github.com/dodopizza/K2Bridge/blob/e75b6986a94a7844b9c090f0550203d7bb8744a5/K2Bridge/KustoDAL/KustoDataAccess.cs#L146

We found out that some of our logs have "*" as property name.

For example, K2Bridge logs such jsons:

```json
{
	"SourceContext": "K2Bridge.KustoDAL.KustoQueryExecutor",
	"queryData": {
		"Data": {
			"$type": "QueryData",
			"QueryCommandText": "REDACTED",
			"HighlightPostTag": "@/kibana-highlighted-field@",
			"HighlightPreTag": "@kibana-highlighted-field@",
			"IndexName": "REDACTED",
			"DocValueFields": [
				"Timestamp"
			],
			"HighlightText": {
				"Cluster": "dev",
				"*": "App:(\"REDACTED\", \"REDACTED\", \"REDACTED\", \"REDACTED\")"
			},
			"SortFields": [
				"Timestamp"
			]
		}
	}
}
```

As you can see here, we have a property `*` under `queryData.Data.HighlightText`. When `bag_unpack` tries to unpack it into a column, it fails. We have to somehow tell bag_unpack not to unpack this into column, but I had a hard time filtering property with such name, it seems that  only `mv-apply` will work...

We also fixed doublequotes escape(#26) because now the filtering works this is a much bigger problem.

We understand that this is a very, very suboptimal and a hacky way to implement this, and are looking for a feedback and ways to improve this. Our approach utilizes sampling, so not 100% of properties get sampled and we do not discover 100% nested properties.

### DIFF
--- a/K2Bridge/KustoDAL/KustoDataAccess.cs
+++ b/K2Bridge/KustoDAL/KustoDataAccess.cs
@@ -19,6 +19,11 @@ namespace K2Bridge.KustoDAL
     /// </summary>
     internal class KustoDataAccess : IKustoDataAccess
     {
+        private const int MaxFieldCapsSampleSize = 8;
+        private const string MaxFieldCapsLookbackTime = "4h";
+        private const int MaxFieldCapsRecursionDepth = 4;
+        private const string MaxFieldCapsUniqueGetSchemaColumn = "K2BridgeGetSchemaColumn";
+
         /// <summary>
         /// Initializes a new instance of the <see cref="KustoDataAccess"/> class.
         /// </summary>
@@ -53,7 +58,8 @@ namespace K2Bridge.KustoDAL
                 string kustoCommand = $".show {KustoQLOperators.Databases} {KustoQLOperators.Schema} | {KustoQLOperators.Where} TableName=='{tableName}' {KustoQLOperators.And} DatabaseName=='{databaseName}' {KustoQLOperators.And} ColumnName!='' | {KustoQLOperators.Project} ColumnName, ColumnType";
 
                 using IDataReader kustoResults = await Kusto.ExecuteControlCommandAsync(kustoCommand, RequestContext);
-                MapFieldCaps(kustoResults, response);
+
+                await MapFieldCapsAsync(kustoResults, response, tableName);
 
                 if (response.Fields.Count > 0)
                 {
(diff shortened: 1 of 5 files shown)


spacentropy:
@boazsha @guregini @eladiw @tamirkamara Please look at this :-)

boazsha:
Hi @spacentropy!
Thanks again for contributing :)
We are looking into this... In the meantime, it seems that some tests fail. Could you please fix them?

spacentropy:
@boazsha sure, will fix them

boazsha:
@spacentropy thanks!
A question: in the past - when we've considered implementing this feature - we thought about using [buildschema()](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/buildschema-aggfunction) instead of getschema().
This works on dynamic columns without the need to make in-depth recursive calls (but an order has to be set on types, in case of contradicting types).
Have you considered this?

spacentropy:
@boazsha sure, we tried to use it first, but it required a more complex changes - it returns one json object with `"property": "type"` schema, which was hard to map to current output used by MapFieldCaps (two columns, ColumnName & ColumnType). `getschema` nicely returns two columns with same types(`System.String` vs `string`), so we've able to quickly make it work.

Also it still wouldn't solve the problem with sampling: imagine if the column has 100TB data...

I can try to move our implementation to buildschema, but it would mean more fundamental changes in `GetFieldCapsAsync`, while we've just changed `MapFieldCaps`. I agree that this would be a better option requiring less queries.

PR #45, opened 2020-09-03
Cache Intermediary Results Using Materialize

boazsha:

* Speed up query time


spacentropy:
@boazsha what if we take 7 days of data, will this materialized `_data` fit in memory?
Now it works because `aggs` query has `summarize` in it, and `hits` query takes only 500 rows.

tamirkamara:
We actually had materialize in the initial versions of the project, however it doesn't hold up in reality - 
_data includes ALL results that match the filters which means it's likely materialize will fail as the base table gets larger (not even huge).

boazsha:
Hi guys, this is indeed the concern. We'll try to make it work... (this PR is mainly for E2E).

PR #48, opened 2020-09-08
Reorder Query Filters per Operator Set

boazsha:

* Query filters which are on dynamic columns will be pushed to the end of the Operator set, which will improve query performance.
* In a future PR, this will be done on a query level instead on a parameter-set level.


K2Bridge/Visitors/BoolClauseVisitor.cs
```diff
@@ -15,44 +17,48 @@ namespace K2Bridge.Visitors
...
-                {
-                    boolQuery.KustoQL += $"{joinString}";
-                }
-
-                // Now, join the given list separated with the given word
-                boolQuery.KustoQL += $"{string.Join(joinString, queryList)}";
+                kustoQuery.Append($"{paddedJoinString}");
             }
+
+            var orderedList = queryList.OrderBy(exp => DynamicRegex.Match(exp).Groups[2].Success);
```
tamirkamara, line 53: I think you should a comment here explaining what this does and why.

tamirkamara (approved):

PR #50, opened 2020-09-15
Fix doublequotes escaping in filters

### LINKED ISSUE #26: Doublequotes escaping is broken

spacentropy:
Fixes #26


K2Bridge/Visitors/QueryStringClauseVisitor.cs
```diff
@@ -67,7 +67,7 @@ public void Visit(QueryStringClause queryStringClause)
                     break;
 
                 case QueryStringClause.Subtype.Phrase:
-                    queryStringClause.KustoQL = $"{queryStringClause.ParsedFieldName} {KustoQLOperators.Contains} \"{queryStringClause.Phrase.EscapeSlashes()}\"";
+                    queryStringClause.KustoQL = $"{queryStringClause.ParsedFieldName} {KustoQLOperators.Contains} \"{queryStringClause.Phrase}\"";
```
tamirkamara, line 70: Isn't it required to escape here (why was it removed)?

spacentropy: This method works when you enter your filter in "query" bar:
image
So, if we want to search for `Coats with "quotes"`, we have to escape doublequote **inside** Kibana query bar, because doublequotes are used by Kibana to mark the boundaries of phrases (like -field:"abc").
This `EscapeSlashes` method escapes the kibana doublequote escape and the query fails:
image

Without these methods those slashes escaping the doublequotes are handed to Kusto and it works.

While writing this I understood that we also have to remove EscapeSlashes at line 64 so `QueryStringClause.Subtype.Term` wouldn't be also double escaped. And on line 85 so `hasprefix` would also work with doublequotes:
image

Maybe I miss some other case where we actually need escaping slashes inside kibana query bar?

tamirkamara: Thank you @spacentropy for these details. 
I understand what you aim to accomplish here but my concern is that your change reverses something else, specifically how to handle _phrases_ that include a **\** char.  taking your example, this phrase will now fail: "Coats with \slash\" (or not return the right result).
I'm sure there's a way to fix that too, but before we go that route - keeping K2 behavior on-par with elastic is/was really hard. Can you share a comparable example from Kibana working with Elaticsearch directly?

PR #55, opened 2021-06-29
Log Analytics queries - Add a note about adding the service principal to the Log Analytics workspace

guregini:
- Add a note about adding the service principal to the Log Analytics workspace
- Set "generic" variables in the function that query Log Analytics


AsafMah (approved):
Looks good, do you think the function name should be generic too?

PR #60, opened 2021-10-27
feat: Add development container for Visual Studio Code

dupuyjs:
# PR Request

Add development container with recommended tools to build and test K2Bridge.
This container is based on Ubuntu 20.10 image. It creates and runs the container with a non-root user named 'dev' (needed as ElasticSearch cannot start with a root user).

ElasticSearch and Kibana are installed in /home/dev.

## Tools
- [x] Terraform 1.0.9
- [x] Azure CLI (latest)
- [x] Helm (latest)
- [x] .NET SDK 3.1
- [x] .NET SDK 5.0

## Elastic Stack
- [x] Elastic Search 6.8.20
- [x] Kibana 6.8.20

And add development container documentation.


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+# Install Helm
+RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
+RUN chmod 700 get_helm.sh
+RUN ./get_helm.sh
+
+# Install .NET SDK
+RUN wget https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
+RUN dpkg -i packages-microsoft-prod.deb
+RUN rm packages-microsoft-prod.deb
```
eladiw, line 39: These lines can probably be removed if using a devcontainer image

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,67 @@
+FROM ubuntu:20.10
```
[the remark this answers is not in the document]
dupuyjs: Updated base image to mcr.microsoft.com/vscode/devcontainers/base:0-ubuntu20.04. It removed installation of some common tools (sudo, git, ..) and creation of a default non-root user, we can use vscode user instead.

.devcontainer/Dockerfile
dupuyjs: Only if we use [dotnet base image](https://github.com/microsoft/vscode-dev-containers/tree/main/containers/dotnet). Inconvenience, you have to select 3.1 or 5.0, not both are installed by the base image. 

As this is not really time consuming, I prefer to use the default dev container image (base:0-ubuntu20.04) and continue to install both SDK in addition. But I am open to change this point if you consider this is important.

eladiw (approved):

dupuyjs:
I updated the PR with documentation for Visual Studio code and a script to deploy ADX + Service Principal for dev purpose. @eladiw @tomconte

.devcontainer/create-adx.sh
```diff
@@ -0,0 +1,66 @@
...
+az extension add -n kusto
+az kusto cluster create --cluster-name $ADX_CLUSTER_NAME --sku name="Dev(No SLA)_Standard_D11_v2" tier="Basic" capacity=1 --resource-group $RESOURCE_GROUP_NAME --location $RESOURCE_GROUP_LOCATION --type="None" 
+
+kusto_scope=$(az kusto cluster show --cluster-name $ADX_CLUSTER_NAME --resource-group $RESOURCE_GROUP_NAME --query id -o tsv)
+echo $kusto_scope
+
+# Create Data Explorer Database
+az kusto database create --cluster-name $ADX_CLUSTER_NAME --database-name $ADX_DB_NAME --resource-group $RESOURCE_GROUP_NAME --read-write-database soft-delete-period=P365D hot-cache-period=P31D location=$RESOURCE_GROUP_LOCATION
+
+# Create Service Pincipal
```
eladiw, line 51:
```suggestion
# Create Service Principal
```

eladiw:
Looks good. If you already created the create ADX script. 
Would it make sense to populate it with data? otherwise the ADX is not so useful and the developer will need to go and research how to do it.

.devcontainer/create-adx.sh
dupuyjs: Fixed.

dupuyjs:
> Looks good. If you already created the create ADX script. Would it make sense to populate it with data? otherwise the ADX is not so useful and the developer will need to go and research how to do it.

Yes. I suppose the best way to achieve this : 1/ populate environment variables at the end of the script needed to run end-to-end tests. 2/ run end-to-end tests to populate data. 

Otherwise, I don't think we can directly populate data with az kusto commands. Do you have any other option in mind ? @eladiw @tomconte

tomconte (approved):
Works for me! 👍🏻

eladiw (approved):

PR #62, opened 2021-11-08
Adding Kibana 7.10 support (instead of Kibana 6.8)

tomconte:

* Handle the /_resolve/index method used by Kibana 7 
* Update the end-to-end tests for Kibana 7 / Elasticsearch 7


K2Bridge/Controllers/IndexListController.cs
```diff
@@ -13,6 +13,8 @@ namespace K2Bridge.Controllers
     /// Handles Index List requests.
     /// The original request produced by Kibana is in the format of:
     /// POST /*/_search?ignore_unavailable=true HTTP/1.1.
+    /// In addition, for Kibana 7:
```
eladiw, line 16: I think 'old' api calls should be removed from the code, meaning lets not try to aim for both Kibana 6.8 and 7.

We can either removed those calls in this PR.
or open a new task to removed redundant calls.
wdyt

K2Bridge/KustoDAL/KustoDataAccess.cs
```diff
@@ -115,6 +115,41 @@ public async Task<IndexListResponseElement> GetIndexListAsync(string indexName)
             return response;
         }
 
+        /// <summary>
+        /// Executes a query to Kusto for Index List.
+        /// Searches for all tables and all functions that match the index name pattern.
+        /// </summary>
+        /// <param name="indexName">Index name pattern, e.g. "*", "orders*", "orders".</param>
+        /// <returns>A list of Indexes matching the given name pattern.</returns>
+        public async Task<ResolveIndexResponse> ResolveIndexAsync(string indexName)
```
eladiw, line 124: Can we delete GetIndexListAsync and every non needed method?
I think its the same exact code, (except one line inside the if statement)

K2Bridge/Controllers/IndexListController.cs
tomconte: Completely agree, however in this case, Kibana 7 *also* sends a `/_search` request, so both `_resolve` and `_search` seem necessary.

K2Bridge/KustoDAL/KustoDataAccess.cs
tomconte: I will look at combining both methods into one.

tomconte: I have refactored to avoid code duplication.

K2Bridge.Tests.UnitTests/Controllers/IndexListControllerTests.cs
```diff
@@ -38,6 +51,7 @@ private IndexListController GetController()
             var responseTask = Task.FromResult(response);
             mockDAL.Setup(kusto => kusto.GetFieldCapsAsync(It.IsNotNull<string>())).Returns(responseTask);
             mockDAL.Setup(kusto => kusto.GetIndexListAsync(It.IsNotNull<string>())).Returns(Task.FromResult(new IndexListResponseElement() { }));
+            mockDAL.Setup(kusto => kusto.ResolveIndexAsync(It.IsNotNull<string>())).Returns(Task.FromResult(new ResolveIndexResponse() { }));
```
eladiw, line 54: Isn't line 53 and its matching method (GetIndexListAsync) redundant?

K2Bridge/Startup.cs
```diff
@@ -156,6 +156,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
                 // Special treatment for certains requests that are intentionally not marked with the [ApiController] attribute.
                 endpoints.MapControllerRoute("fieldcaps", "FieldCapability/Process/{indexName?}", defaults: new { controller = "FieldCapability", action = "Process" });
                 endpoints.MapControllerRoute("indexlist", "IndexList/Process/{indexName?}", defaults: new { controller = "IndexList", action = "Process" });
```
eladiw, line 158: Is this line still needed? (and the Process method itself)

K2Bridge/Controllers/IndexListController.cs
```diff
@@ -43,5 +45,18 @@ public async Task<IActionResult> Process(string indexName)
```
eladiw, line 45: This method can be deleted:
public async Task<IActionResult> Process(string indexName) 

(verified - works without it)

K2Bridge/Startup.cs
```diff
@@ -156,6 +156,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+                endpoints.MapControllerRoute("resolveindex", "IndexList/Resolve/{indexName?}", defaults: new { controller = "IndexList", action = "Resolve" });
```
eladiw, line 159: This line can be removed:

endpoints.MapControllerRoute("indexlist", "IndexList/Process/{indexName?}", defaults: new { controller = "IndexList", action = "Process" });

K2Bridge/KustoDAL/IKustoDataAccess.cs
```diff
@@ -25,5 +25,12 @@ public interface IKustoDataAccess
         /// <param name="indexName">Index name.</param>
         /// <returns>A list of Indexes.</returns>
         Task<IndexListResponseElement> GetIndexListAsync(string indexName);
```
eladiw, line 27: This method can be removed from the interface and the implementation

K2Bridge/RewriteRules/RewriteSearchRule.cs
```diff
@@ -65,12 +67,23 @@ public async void ApplyRule(RewriteContext context)
                     context.HttpContext.Request.Path = $"/Query/SingleSearch/{GetIndexNameFromPath(context.HttpContext.Request.Path)}";
                 }
```
eladiw, line 68: L57-63 (and line 66) in the original file, can be removed ( I think its not used)

eladiw (changes requested):
Looks good!
I played with it and it works well. I managed to delete some redundant code (of Kibana 6.8) and it still worked well.

K2Bridge.Tests.UnitTests/Controllers/IndexListControllerTests.cs
tomconte: Removed.

K2Bridge/Startup.cs
tomconte: Removed.

K2Bridge/RewriteRules/RewriteSearchRule.cs
```diff
@@ -65,12 +67,23 @@ public async void ApplyRule(RewriteContext context)
...
             }
+            else if (context.HttpContext.Request.Path.Value.Contains("_resolve/index", System.StringComparison.OrdinalIgnoreCase))
+            {
+                // Resolve index query from Kibana 7
+                context.HttpContext.Request.Path = $"/IndexList/Resolve/{GetIndexNameFromResolvePath(context.HttpContext.Request.Path)}";
+            }
         }
 
         private string GetIndexNameFromPath(PathString pathString)
```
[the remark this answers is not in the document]
tomconte: Removed.

K2Bridge/Controllers/IndexListController.cs
tomconte: Removed.

K2Bridge/Startup.cs
tomconte: Removed.

K2Bridge/KustoDAL/IKustoDataAccess.cs
tomconte: Removed.

K2Bridge/RewriteRules/RewriteSearchRule.cs
tomconte: Removed.

K2Bridge/KustoDAL/KustoDataAccess.cs
```diff
@@ -93,7 +94,10 @@ public async Task<IndexListResponseElement> GetIndexListAsync(string indexName)
                 using IDataReader kustoTables = await Kusto.ExecuteControlCommandAsync(readTablesCommand, RequestContext);
                 if (kustoTables != null)
                 {
-                    MapIndexList(kustoTables, response);
+                    while (kustoTables.Read())
+                    {
+                        tablesAndFunctions.Add(Convert.ToString(kustoTables[0]));
```
eladiw, line 99: Not sure what happens if there are no tables, exception?

K2Bridge/KustoDAL/KustoDataAccess.cs
```diff
@@ -103,12 +107,38 @@ public async Task<IndexListResponseElement> GetIndexListAsync(string indexName)
                 using IDataReader kustoFunctions = await Kusto.ExecuteControlCommandAsync(readFunctionsCommand, RequestContext);
                 if (kustoFunctions != null)
                 {
-                    MapIndexList(kustoFunctions, response);
+                    while (kustoFunctions.Read())
+                    {
+                        tablesAndFunctions.Add(Convert.ToString(kustoFunctions[0]));
```
eladiw, line 112: what happens if there are no functions? always the error below will be thrown?

eladiw (commented):
I think that the 'old' IndexListResponseElement class can be deleted

K2Bridge/KustoDAL/KustoDataAccess.cs
tomconte: No exception if there are no rows to read, `Read()` will return false and no element will be added to the list.

K2Bridge/KustoDAL/KustoDataAccess.cs
tomconte: Same as above, no element will be added, possibly ending up with an empty list and empty response. I could add some unit tests if you think that is necessary (I don't think these cases are covered by current tests).

K2Bridge/RewriteRules/RewriteSearchRule.cs
```diff
@@ -72,5 +41,11 @@ private string GetIndexNameFromPath(PathString pathString)
             var segments = pathString.ToString().Split('/', System.StringSplitOptions.RemoveEmptyEntries);
             return segments[0];
         }
+
+        private string GetIndexNameFromResolvePath(PathString pathString)
```
tamirkamara, line 45: Can you extract the index search pattern implicitly by decorating the controller method instead of using this method explicitly?

K2Bridge/RewriteRules/RewriteSearchRule.cs
```diff
@@ -18,52 +18,21 @@ internal class RewriteSearchRule : IRule
...
-                {
-                    // This is a regular search (documents) request
-                    context.HttpContext.Request.Path = $"/Query/SingleSearch/{GetIndexNameFromPath(context.HttpContext.Request.Path)}";
-                }
+                context.HttpContext.Request.Path = $"/Query/SingleSearch/{GetIndexNameFromPath(context.HttpContext.Request.Path)}";
+            }
+            else if (context.HttpContext.Request.Path.Value.Contains("_resolve/index", System.StringComparison.OrdinalIgnoreCase))
+            {
+                // Resolve index query from Kibana 7
+                context.HttpContext.Request.Path = $"/IndexList/Resolve/{GetIndexNameFromResolvePath(context.HttpContext.Request.Path)}";
```
tamirkamara, line 35: Since the route pattern has changed and became simpler in the sense it's not a subset of the parent _search route, it looks like we can remove rewrites and have the route definitions directly on the controller. Can you please have another look at this?

K2Bridge/Controllers/IndexListController.cs
```diff
@@ -32,14 +32,14 @@ public IndexListController(IKustoDataAccess kustoDataAcess, ILogger<IndexListCon
...
         /// <summary>
-        /// Process the request.
+        /// Resolve the index.
         /// </summary>
-        /// <param name="indexName">The index to process.</param>
+        /// <param name="indexName">The index pattern to process.</param>
         /// <returns>The table list in the Kusto database.</returns>
         [Produces("application/json")]
-        public async Task<IActionResult> Process(string indexName)
+        public async Task<IActionResult> Resolve(string indexName)
```
tamirkamara, line 40: See my comment below, I think you can have the routing logic applied here as decoration.

tamirkamara (changes requested):
Nice @tomconte. I had just one comment regarding an option to remove the rewrite rule (if I haven't missing something).
Also, Can you please rename the title of this PR to reflect it being focused on the index list?

K2Bridge.Tests.End2End/KustoFunctionsTests.cs
```diff
@@ -43,6 +43,9 @@ public async Task FieldCaps_WithFunction_ReturnsResponseWithFunctions()
         {
             var fieldCaps = await K2Client().FieldCaps(functionFullName);
             var expected = JObject.Parse(@"{
+              ""indices"": [
+                ""fn_countries_and_airports""
```
eladiw, line 47: you can  use functionFullName
(which  is in the form of  db:fn_name, so maybe take the last part.

K2Bridge.Tests.End2End/KustoClientTests.cs
```diff
@@ -91,6 +91,9 @@ public async Task FieldCaps_WithAllTypes_ReturnsTypesConvertedToESTypes()
         {
             var fieldCaps = await K2Client().FieldCaps(typesIndexFullName);
             var expected = JObject.Parse(@"{
+              ""indices"": [
+                ""types_index""
```
eladiw, line 95: can  usetypes IndexFullName

(need to take substring)

K2Bridge.Tests.End2End/TestElasticClient.cs
```diff
@@ -253,8 +232,11 @@ private static void NormalizeIndexNamesForIndexList(JToken parent, string jsonPa
             {
                 if (tokens[index].Value.Equals(indexName))
                 {
-                    var tableName = indexName.Split(':')[1];
-                    tokens[index].Value = tableName;
+                    if (indexName.Contains(':'))
```
eladiw, line 235: I think that  you are correct and the ':' must exists. did you made sure that its the case?

K2Bridge.Tests.End2End/flights/MSearch_Invalid_Query.json
```diff
@@ -9,14 +9,11 @@
...
   ],
-  "_source": {
-    "excludes": []
-  },
   "aggs": {
     "2": {
       "date_histogram": {
         "field": "timestamp",
-        "interval": "1M",
+        "calendar_interval": "month",
```
eladiw, line 16: for other reviewers.
I made sure to have both calendar_interval and fixed_interval as both of these replace the old 'interval'. making sure we  support both

K2Bridge.Tests.End2End/flights/MSearch_MultipleFilters_Equivalent.json
```diff
@@ -25,23 +22,35 @@
...
   ],
-  "script_fields": {},
+  "script_fields": {
+    "hour_of_day": {
+      "script": {
+        "source": "doc['timestamp'].value.hourOfDay",
+        "lang": "painless"
+      }
+    }
+  },
```
eladiw, line 32: generated by Kibana

K2Bridge.Tests.UnitTests/Controllers/QueryControllerTests.cs
```diff
@@ -24,10 +24,10 @@ namespace UnitTests.K2Bridge.Controllers
...
+        private const string ValidMSearchRequestContent = "{\"index\":\"kibana_sample_data_flights\",\"ignore_unavailable\":true,\"preference\":1572955935509}\n{\"version\":true,\"size\":500,\"sort\":[{\"timestamp\":{\"order\":\"desc\",\"unmapped_type\":\"boolean\"}}],\"_source\":{\"excludes\":[]},\"aggs\":{\"2\":{\"date_histogram\":{\"field\":\"timestamp\",\"fixed_interval\":\"1d\",\"time_zone\":\"America/Los_Angeles\",\"min_doc_count\":1}}},\"stored_fields\":[\"*\"],\"script_fields\":{},\"docvalue_fields\":[{\"field\":\"timestamp\",\"format\":\"date_time\"}],\"query\":{\"bool\":{\"must\":[{\"match_all\":{}},{\"range\":{\"timestamp\":{\"gte\":1561673881638,\"lte\":1566712210749,\"format\":\"epoch_millis\"}}}],\"filter\":[],\"should\":[],\"must_not\":[]}},\"highlight\":{\"pre_tags\":[\"@kibana-highlighted-field@\"],\"post_tags\":[\"@/kibana-highlighted-field@\"],\"fields\":{\"*\":{}},\"fragment_size\":2147483647},\"timeout\":\"30000ms\"}";
```
eladiw, line 27: Need to change.its not enough to change the 'interval' to 'fixed_interval', since the query changed

eladiw: (I commented on this although I did it (: )

K2Bridge.Tests.UnitTests/KustoDAL/ElasticResponseTests.cs
```diff
@@ -103,7 +103,7 @@ public string MapRowsToHits_WhenMultiWordDifferentCapsCapialLettersInQueryString
         }
 
         [TestCase(ExpectedResult =
-            "{\"responses\":[{\"aggregations\":{\"2\":{\"buckets\":[]}},\"took\":0,\"timed_out\":false,\"_shards\":{\"total\":1,\"successful\":1,\"skipped\":0,\"failed\":0},\"hits\":{\"total\":0,\"max_score\":null,\"hits\":[]},\"status\":200}]}")]
+            "{\"responses\":[{\"aggregations\":{\"2\":{\"buckets\":[]}},\"took\":0,\"timed_out\":false,\"_shards\":{\"total\":1,\"successful\":1,\"skipped\":0,\"failed\":0},\"hits\":{\"total\":{\"relation\":\"eq\",\"value\":0},\"max_score\":null,\"hits\":[]},\"status\":200}],\"took\":0}")]
```
eladiw, line 106: is this the actual captured response ?

K2Bridge/JsonConverters/DateHistogramAggregationConverter.cs
```diff
@@ -23,10 +23,14 @@ internal class DateHistogramAggregationConverter : ReadOnlyJsonConverter
...
 
+            var fixedInterval = jo["fixed_interval"];
+            var calendarInterval = jo["calendar_interval"];
+            var interval =
+                fixedInterval != null ? (string)fixedInterval : (string)calendarInterval;
             var obj = new DateHistogramAggregation
             {
                 FieldName = (string)jo["field"],
-                Interval = (string)jo["interval"],
+                Interval = interval,
```
eladiw, line 33: explain: Interval can be 1M, month, 1d, day, week, 2d, etc
but the key can be either fixed_interval of calendar_interval, so the logic in this file gets the correct value from either key

eladiw (changes requested):
Commented

K2Bridge/RewriteRules/RewriteSearchRule.cs
tomconte: Thanks, Tamir, done.

K2Bridge/RewriteRules/RewriteSearchRule.cs
tomconte: 👍🏻

K2Bridge/Controllers/IndexListController.cs
tomconte: 👍🏻

K2Bridge.Tests.UnitTests/KustoDAL/ElasticResponseTests.cs
tomconte: No, I just modified the JSON to match the updated response format ("total" is an object, new "took" element at top-level).

eladiw:
Another change needed: change the helm chart to install elastic 7.10.2 to match the expected Kibana version

K2Bridge.Tests.End2End/KustoFunctionsTests.cs
tomconte: 👍🏻

K2Bridge.Tests.End2End/KustoClientTests.cs
tomconte: 👍🏻

K2Bridge.Tests.End2End/TestElasticClient.cs
tomconte: I am now calling this function also on the Elastic results (to prune unwanted indices), in which case the index name has no ':' sign.

https://github.com/microsoft/K2Bridge/blob/feature/kibana-7-resolve-index/K2Bridge.Tests.End2End/ParallelApiTest.cs#L169

K2Bridge.Tests.End2End/KustoClientTests.cs
```diff
@@ -90,80 +90,83 @@ public async Task MSearch_All_ReturnsAllHitsAsExpected()
...
                     ""aggregatable"": true,
                     ""searchable"": true,
                     ""type"": ""date""
-                  }
-                },
-                ""Guid"": {
-                  ""string"": {
+                  }}
+                }},
+                ""Guid"": {{
```
eladiw, line 112: why the extra '{' on each line

K2Bridge.Tests.End2End/ParallelApiTest.cs
```diff
@@ -163,10 +163,10 @@ public void CompareElasticKusto_WhenMSearchSort_ResponsesAreEquivalent(string qu
         }
 
         [Test]
-        [Description("/_search index list Kibana query")]
+        [Description("/_resolve/index Kibana query")]
         public void CompareElasticKusto_WhenSearch_ResponsesAreEquivalent()
         {
-            var es = ESClient().Search();
+            var es = ESClient().Search($"{INDEX}");
```
eladiw, line 169: so now you must pass an index?

K2Bridge.Tests.End2End/TestElasticClient.cs
```diff
@@ -139,47 +141,24 @@ public async Task<JToken> MSearch(string indexName, string jsonQueryFile, bool v
...
 
         /// <summary>
-        /// API operation for wildcard index search (hitting the IndexList endpoint).
+        /// API operation for resolving indices (hitting the IndexList endpoint).
         /// </summary>
         /// <param name="optionalIndexToKeep">Optional input with index name to keep.
         /// if this is not null, all other index names will be removed and it will be
         /// normalized by removing the database name from kusto's db:table pair.</param>
         /// <returns><c>JToken</c> with parsed response.</returns>
         public async Task<JToken> Search(string optionalIndexToKeep = null)
```
eladiw, line 150: looks like its still optional (the index to keep). so why by default you pass the 'index' in the e2e test (commented there also)

K2Bridge/RewriteRules/RewriteSearchRule.cs
```diff
@@ -18,52 +18,15 @@ internal class RewriteSearchRule : IRule
...
-                {
-                    // This is a request for the index list
-                    context.HttpContext.Request.Path = $"/IndexList/Process/{GetIndexNameFromPath(context.HttpContext.Request.Path)}";
-                }
-                else
-                {
-                    // This is a regular search (documents) request
-                    context.HttpContext.Request.Path = $"/Query/SingleSearch/{GetIndexNameFromPath(context.HttpContext.Request.Path)}";
-                }
+                context.HttpContext.Request.Path = $"/Query/SingleSearch/{GetIndexNameFromPath(context.HttpContext.Request.Path)}";
```
eladiw, line 29: can you briefly explain this change?

K2Bridge.Tests.End2End/KustoClientTests.cs
tomconte: Needed to escape the braces characters in interpolated strings:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated

K2Bridge.Tests.End2End/ParallelApiTest.cs
tomconte: So that extra indices (like the kibana index) can be removed from the ES results.
I now realize that this index only exists if Kibana has been launched, which is not normally the case in end-to-end tests... But it was the case on my machine :) So maybe this is not required indeed.

K2Bridge.Tests.End2End/TestElasticClient.cs
tomconte: Yes, it should not be optional if we keep passing the index to ES also.

K2Bridge/RewriteRules/RewriteSearchRule.cs
tomconte: Following Tamir's comment, I removed the rewrite rule completely for _resolve/index, since the routing can be implemented using regular decorations (in IndexListController). All that is left is the rewrite rule for _search queries.

K2Bridge.Tests.End2End/ParallelApiTest.cs
tomconte: I have reverted this change; it works fine as long as Kibana is not launched before running the end-to-end tests.

K2Bridge.Tests.End2End/TestElasticClient.cs
tomconte: Reverted (see above).

K2Bridge.Tests.End2End/TestElasticClient.cs
tomconte: Not needed anymore (removed the other changes).

K2Bridge.Tests.UnitTests/Controllers/IndexListControllerTests.cs
AsafMah: Why was this removed? do we know that kibana 7 doesn't request the index list in this way?

tomconte: Kibana 7 uses the `/_resolve/index` API and we did not want to keep the 6.x code around.

https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-resolve-index-api.html

K2Bridge/Visitors/DateHistogramVisitor.cs
```diff
@@ -23,13 +23,26 @@ public void Visit(DateHistogramAggregation dateHistogramAggregation)
             if (!string.IsNullOrEmpty(dateHistogramAggregation.Interval))
             {
                 var period = dateHistogramAggregation.Interval[^1];
-                dateHistogramAggregation.KustoQL += period switch
+                if (dateHistogramAggregation.Interval.Contains("week", System.StringComparison.OrdinalIgnoreCase))
```
AsafMah, line 26: What do you think about this simplification:

```c#
var interval = dateHistogramAggregation.Interval;
                var period = interval[^1];
                dateHistogramAggregation.KustoQL += period switch
                {
                    'w' => $"{KustoQLOperators.StartOfWeek}({dateHistogramAggregation.FieldName})",
                    'M' => $"{KustoQLOperators.StartOfMonth}({dateHistogramAggregation.FieldName})",
                    'y' => $"{KustoQLOperators.StartOfYear}({dateHistogramAggregation.FieldName})",
                    _ when interval.Contains("week", System.StringComparison.OrdinalIgnoreCase) => $"{KustoQLOperators.StartOfWeek}({dateHistogramAggregation.FieldName})",
                    _ when interval.Contains("month", System.StringComparison.OrdinalIgnoreCase) => $"{KustoQLOperators.StartOfMonth}({dateHistogramAggregation.FieldName})",
                    _ when interval.Contains("year", System.StringComparison.OrdinalIgnoreCase) => $"{KustoQLOperators.StartOfYear}({dateHistogramAggregation.FieldName})",
                    _ => $"bin({dateHistogramAggregation.FieldName}, {dateHistogramAggregation.Interval})",
                };
```

K2Bridge.Tests.UnitTests/Controllers/IndexListControllerTests.cs
AsafMah: What about this comment from earlier -https://github.com/microsoft/K2Bridge/pull/62#discussion_r744722431?

tomconte: Kibana 7 does send a `/*%3A*/_search` request but it does not seem to be used. Index patterns management work fine without it.

tomconte: Below is a link to the source code of Kibana 7.10.2 where indices are searched for using `_search`:

https://github.com/elastic/kibana/blob/v7.10.2/src/plugins/index_pattern_management/public/components/create_index_pattern_wizard/lib/get_indices.ts#L199

According to the comment, "// CCS supports ±1 major version. We won't be able to expect resolve endpoint to exist until v9". 

So, my understanding is that the `_search` call is there only to supports Cross Cluster Search (CCS), which does not impact us.

eladiw: good discussion.

AsafMah (approved):

PR #63, opened 2021-11-08
Fix End2End tests for Elasticsearch 7

tomconte:

* Changes to make End2End tests pass on Elasticsearch 7.


tomconte:
All these changes have been merged into #62.

PR #64, opened 2021-11-10
Updates to the e2e queries - Kibana 7(.10.2)

eladiw:

* Kept the same logic in the queries but with the new format of Kibana 7


tomconte (commented):
Should we have a mix of fixed_interval and calendar_interval in the queries so we cover both cases?

tomconte (approved):
Looks good to me!

AsafMah:
How did you do these changes? 
Manually? Find + replace? some tool?

eladiw:
no. captured new queries.
one note: the timestamps are manual because I wanted it to be identical.

PR #66, opened 2021-11-16
Upgrading to .net 6.0

eladiw:

* Upgrading the .net version to latest (6.0)
* Had to remove the use of WebClient which is obsolete since .net 6

### DIFF
--- a/K2Bridge.Tests.End2End/KustoTestBase.cs
+++ b/K2Bridge.Tests.End2End/KustoTestBase.cs
@@ -120,30 +120,37 @@ namespace K2Bridge.Tests.End2End
             kusto = new KustoConnectionStringBuilder(kustoUri, kustoDatabase)
                 .WithAadApplicationTokenAuthentication(token);
 
-            if (!File.Exists("flights.json.gz"))
+            const string fileName = "flights.json.gz";
+            if (!File.Exists(fileName))
             {
-                using var hc = new HttpClient();
-                var response =
-                    await hc.GetAsync(new Uri(
-                        "https://raw.githubusercontent.com/elastic/kibana/v6.8.5/src/server/sample_data/data_sets/flights/flights.json.gz"));
-                if (response.IsSuccessStatusCode)
-                {
-                    using (var stream = await response.Content.ReadAsStreamAsync())
-                    {
-                        var fi = new FileInfo("flights.json.gz");
-                        using (var fs = fi.OpenWrite())
-                        {
-                            await stream.CopyToAsync(fs);
-                        }
-                    }
-                }
-                else
-                {
-                    throw new FileNotFoundException();
-                }
+                await DownloadFile("https://raw.githubusercontent.com/elastic/kibana/v6.8.5/src/server/sample_data/data_sets/flights/flights.json.gz", fileName);
             }
 
-            await PopulateBothBackends($"{FLIGHTSDIR}/structure.json", "flights.json.gz");
+            await PopulateBothBackends($"{FLIGHTSDIR}/structure.json", fileName);
+        }
+
+        /// <summary>
+        /// Download a remote file.
+        /// </summary>
+        /// <param name="url">the URL of the remote file</param>
+        /// <param name="fileName">The file to be created</param>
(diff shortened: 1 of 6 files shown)


K2Bridge.Tests.End2End/KustoTestBase.cs
```diff
@@ -118,12 +118,29 @@ public static async Task Init()
...
 
             kusto = new KustoConnectionStringBuilder(kustoUri, kustoDatabase)
-               .WithAadApplicationTokenAuthentication(token);
+                .WithAadApplicationTokenAuthentication(token);
 
             if (!File.Exists("flights.json.gz"))
             {
-                using var wc = new WebClient();
-                wc.DownloadFile("https://raw.githubusercontent.com/elastic/kibana/v6.8.5/src/server/sample_data/data_sets/flights/flights.json.gz", "flights.json.gz");
+                using var hc = new HttpClient();
```
AsafMah, line 125: Move this into a function?

K2Bridge.Tests.End2End/KustoTestBase.cs
```diff
@@ -118,12 +118,29 @@ public static async Task Init()
...
+                var response =
+                    await hc.GetAsync(new Uri(
+                        "https://raw.githubusercontent.com/elastic/kibana/v6.8.5/src/server/sample_data/data_sets/flights/flights.json.gz"));
+                if (response.IsSuccessStatusCode)
+                {
+                    using (var stream = await response.Content.ReadAsStreamAsync())
+                    {
+                        var fi = new FileInfo("flights.json.gz");
+                        using (var fs = fi.OpenWrite())
```
AsafMah, line 134: If you're using the "using declaration" above, you can do it here to:

```c#
                    using var stream = await response.Content.ReadAsStreamAsync();
                    var fi = new FileInfo("flights.json.gz");
                    using var fs = fi.OpenWrite();
                    await stream.CopyToAsync(fs);
```
You can upgrade it further by using "await using", which will dispose asynchronously:
```c#
                    await using var stream = await response.Content.ReadAsStreamAsync();
                    var fi = new FileInfo("flights.json.gz");
                    await using var fs = fi.OpenWrite();
                    await stream.CopyToAsync(fs);
```

Also, is that really the minimal amount of code to achieve the same functionallity as WebClient?

K2Bridge.Tests.End2End/TestElasticClient.cs
```diff
@@ -120,7 +120,7 @@ public async Task<JToken> MSearch(string indexName, string jsonQueryFile, bool v
 
             using var request = new HttpRequestMessage(HttpMethod.Post, "_msearch");
             var payload = new StringBuilder();
-            payload.AppendLine($"{{\"index\":\"{indexName}\"}}");
+            payload.AppendLine(FormattableString.Invariant($"{{\"index\":\"{indexName}\"}}"));
```
AsafMah, line 123: Why is it needed? Why only here?

eladiw: This is the only place the compiler issued a warning

K2Bridge.Tests.End2End/KustoTestBase.cs
eladiw: about your last comment, the answer is yes.

K2Bridge.Tests.End2End/KustoTestBase.cs
```diff
@@ -211,5 +211,29 @@ private static async Task<TestElasticClient> CreateKustoClient(string prefix)
...
+        /// <summary>
+        /// Download a remote file.
+        /// </summary>
+        /// <param name="url">the URL of the remote file</param>
+        /// <param name="fileName">The file to be created</param>
+        /// <exception cref="FileNotFoundException"></exception>
+        private static async Task DownloadFile(string url, string fileName)
+        {
+            using var hc = new HttpClient();
+            var response =
```
AsafMah, line 224: nit: no need for a line break

AsafMah (approved):

PR #67, opened 2021-11-17
feat: Add scripts to provision ADX and ingest sample data (dev environment).

dupuyjs:
# PR Request

Add scripts to provision ADX and ingest sample data (dev environment).
Following steps are performed by this script: 

- [x] Create resource group
- [x] Create Azure Data Explorer cluster
- [x] Create Azure Data Explorer database
- [x] Create service principal
- [x] Assign service principal to database permissions (Viewer role)
- [x] Create Azure Data Explorer table
- [x] Ingest sample data
- [x] Add environment variables to .bashrc (will be used by end-to-end tests)


.devcontainer/adx/create-adx.sh
```diff
@@ -0,0 +1,131 @@
...
+# Create Data Explorer Cluster
+az extension add -n kusto
+echo "$(tput setaf 2)[1/8] Creating Data Explorer Cluster. Please be patient. Long running operation (10 minutes).$(tput setaf 7)"
+az kusto cluster create \
+    --cluster-name $ADX_CLUSTER_NAME \
+    --sku name="Dev(No SLA)_Standard_D11_v2" tier="Basic" capacity=1 \
+    --resource-group $RESOURCE_GROUP_NAME \
+    --location $RESOURCE_GROUP_LOCATION \
+    --enable-streaming-ingest true \
+    --type="None"
```
AsafMah, line 50: What's --type? I don't see it in the doc page

.devcontainer/adx/create-adx.sh
```diff
@@ -0,0 +1,131 @@
...
+az kusto database create \
+    --cluster-name $ADX_CLUSTER_NAME \
+    --database-name $ADX_DB_NAME \
+    --resource-group $RESOURCE_GROUP_NAME \
+    --read-write-database soft-delete-period=P365D hot-cache-period=P31D location=$RESOURCE_GROUP_LOCATION
+
+# Create Service Principal
+echo "$(tput setaf 2)[3/8] Creating Service Principal.$(tput setaf 7)"
+SECRETS=$(az ad sp create-for-rbac --name $SERVICE_PRINCIPAL_NAME --role Contributor --scopes $RESOURCE_GROUP_SCOPE)
+SERVICE_PRINCIPAL_FQN="aadapp=$(echo $SECRETS | jq -r '.appId');$(echo $SECRETS | jq -r '.tenant')"
```
AsafMah, line 63: So this script also depends on `jq`, you should probably mention it in the doc, or avoid it (using -otsv and cut like in other places, though I will admit it's prone to breaking)

.devcontainer/adx/create-adx.sh
```diff
@@ -0,0 +1,131 @@
...
+az storage blob upload --container-name $STORAGE_CONTAINER_NAME --file table.kql
+
+az kusto script create \
+    --cluster-name $ADX_CLUSTER_NAME \
+    --database-name $ADX_DB_NAME \
+    --resource-group $RESOURCE_GROUP_NAME \
+    --script-url "https://$STORAGE_ACCOUNT_NAME.blob.core.windows.net/$STORAGE_CONTAINER_NAME/table.kql" \
+    --script-url-sas-token "?$SAS" \
+    --name "script" \
+    --force-update-tag $(cat /proc/sys/kernel/random/uuid)
```
AsafMah, line 105: So we're creating the storage account and everything just to upload the script?

Since the script doesn't change, can't we just have a direct link to it like with flights.json.gz?

.devcontainer/adx/create-adx.sh
```diff
@@ -0,0 +1,131 @@
...
+TOKEN=$(az account get-access-token --scope "$KUSTO_URL/.default" -o tsv | cut -f1)
+
+curl -X POST "$KUSTO_URL/v1/rest/ingest/$ADX_DB_NAME/kibana_data_flights?streamFormat=json" --header "Authorization: Bearer $TOKEN" --header "Content-Encoding: gzip" --header "Content-Type: multipart/form-data" --data-binary "@$DATA_FILE_NAME"
+rm $DATA_FILE_NAME
+
+# Environment variables needed to run end-to-end tests
+echo "export K2BRIDGE_URL=http://localhost:8080" >> ~/.bashrc 
+echo "export ELASTICSEARCH_URL=http://localhost:9200" >> ~/.bashrc 
+echo "export KUSTO_URI=https://$ADX_CLUSTER_NAME.$RESOURCE_GROUP_LOCATION.kusto.windows.net" >> ~/.bashrc 
+echo "export KUSTO_DB=$ADX_DB_NAME" >> ~/.bashrc
```
AsafMah, line 122: We can also set AAD_TOKEN

AsafMah (changes requested):

.devcontainer/adx/create-adx.sh
dupuyjs: This is an issue reported at https://github.com/Azure/azure-cli/issues/18827. The current workaround is to set this --type= "None"' flag.

.devcontainer/adx/create-adx.sh
dupuyjs: Yes, jq has been installed by default on the dev container. But I take your point and will investigate if I can update this code.

.devcontainer/adx/create-adx.sh
dupuyjs: In local development, can be omitted, the script will then use the Azure CLI to generate a token. (I tested and it works fine)
If I set AAD_TOKEN, it will probably need to be renewed if end-to-end tests are launched 2 days later.

.devcontainer/adx/create-adx.sh
dupuyjs: It's a good point. But if I figured out correctly, 'az kusto script create' takes only script stored from a blob storage. 
I tried to target an Url and get this exception: 
(InvalidScriptProperties) The request is missing the following 'scriptUrlSasToken' properties. Please provide both scriptUrl and scriptUrlSasToken.
Code: InvalidScriptProperties
Message: The request is missing the following 'scriptUrlSasToken' properties. Please provide both scriptUrl and scriptUrlSasToken.

Then, with a dummy SasToken: 
(Failed) Internal Server Error
Code: Failed
Message: Internal Server Error

.devcontainer/adx/create-adx.sh
AsafMah: Fair enough.
So I think that we should still explicitly mention it as a dependency just in case somebody tries that in their own machine (and when I think about it, also the az cli).
And if we're using it we should probably be consistent and change the `-o tsv` we use to jq to be more robust

.devcontainer/adx/create-adx.sh
AsafMah: Ah, got it.
Maybe add a comment in the script with a link to the issue?

.devcontainer/adx/create-adx.sh
AsafMah: Ok I understand, so the way I see it we got 3 options:
1. Keep as is
2. Create a read-only blob with the script on our account, with a SAS that expires in 20 years and hardcode it in here
3. Skip the script entirely, and use kusto's rest api to create the tables:
```
curl -X POST "$KUSTO_URL/v1/rest/mgmt" --header "Content-Type: application/json; charset=utf-8" --header "Authorization: Bearer $TOKEN" --data "{\"db\":\"$DATABASE\",\"csl\":\".create table kibana_data_flights (AvgTicketPrice: real, Cancelled: bool, Carrier: string, Dest: string, DestAirportID: string, DestCityName: string, DestCountry: string, DestLocation: dynamic, DestRegion: string, DestWeather: string, DistanceKilometers: real, DistanceMiles: real, FlightDelay: bool, FlightDelayMin: long, FlightDelayType: string, FlightNum: string, FlightTimeHour: real, FlightTimeMin: real, Origin: string, OriginAirportID: string, OriginCityName: string, OriginCountry: string, OriginLocation: dynamic, OriginRegion: string, OriginWeather: string, dayOfWeek: int, timestamp: datetime)\"}"
curl -X POST "$KUSTO_URL/v1/rest/mgmt" --header "Content-Type: application/json; charset=utf-8" --header "Authorization: Bearer $TOKEN" --data "{\"db\":\"$DATABASE\",\"csl\":\".alter table kibana_data_flights policy streamingingestion enable\"}"
```

What do you think?

.devcontainer/adx/create-adx.sh
```diff
@@ -0,0 +1,131 @@
+#!/bin/bash
+
+# Display Help
+Help()
+{
+   echo "Create data explorer cluster for development purpose."
```
eladiw, line 6:
```suggestion
   echo "Create Azure Data Explorer (ADX) cluster for development purpose."
```

.devcontainer/adx/create-adx.sh
```diff
@@ -0,0 +1,131 @@
...
+    esac
+done
+
+if [ ! "$unique" ]; then
+    echo "Please add unique name argument."
+    exit;
+fi
+
+# Resource Names
+RESOURCE_GROUP_LOCATION="westeurope"
```
eladiw, line 30: 'location' might be a good option to be able to pass as arg

.devcontainer/adx/create-adx.sh
```diff
@@ -0,0 +1,131 @@
...
+STORAGE_ACCOUNT_NAME="stk2$unique"
+STORAGE_CONTAINER_NAME="scripts"
+SERVICE_PRINCIPAL_NAME="sp-k2-$unique"
+
+# Create Resource Group
+RESOURCE_GROUP_SCOPE=$(az group create --location $RESOURCE_GROUP_LOCATION --resource-group $RESOURCE_GROUP_NAME --query id -o tsv)
+
+# Create Data Explorer Cluster
+az extension add -n kusto
+echo "$(tput setaf 2)[1/8] Creating Data Explorer Cluster. Please be patient. Long running operation (10 minutes).$(tput setaf 7)"
```
eladiw, line 43:
```suggestion
echo "$(tput setaf 2)[1/8] Creating Azure Data Explorer Cluster. Please be patient. Long running operation (10 minutes).$(tput setaf 7)"
```

.devcontainer/adx/create-adx.sh
```diff
@@ -0,0 +1,131 @@
...
+SECRETS=$(az ad sp create-for-rbac --name $SERVICE_PRINCIPAL_NAME --role Contributor --scopes $RESOURCE_GROUP_SCOPE)
+SERVICE_PRINCIPAL_FQN="aadapp=$(echo $SECRETS | jq -r '.appId');$(echo $SECRETS | jq -r '.tenant')"
+SERVICE_PRINCIPAL_DISPLAYNAME=$(echo $SECRETS | jq '.displayName')
+
+# Wait Service Principal propagation
+echo "$(tput setaf 2)[4/8] Wait 60 seconds until Service Principal is propagated.$(tput setaf 7)"
+sleep 60
+
+# Assign Service Principal to database permissions
+echo "$(tput setaf 2)[5/8] Assigning database permissions.$(tput setaf 7)"
```
eladiw, line 71: I would output the SP id, so the user knows which SP to delete when done with the env.
(also can print a reminder message about deleting it)

.devcontainer/adx/create-adx.sh
eladiw: I vote 1 or 3.

docs/devcontainer-adx.md
```diff
@@ -0,0 +1,58 @@
...
+```bash
+vscode ➜ /workspaces/K2Bridge (feature/script-adx-install ✗) $ az account show
+```
+
+### Change directory to .devcontainer/adx
+```bash
+vscode ➜ /workspaces/K2Bridge (feature/script-adx-install ✗) $ cd ./.devcontainer/adx
+```
+
+### Execute create-adx.sh with a unique name.
```
eladiw, line 37: add location to the args

docs/devcontainer-adx.md
```diff
@@ -0,0 +1,58 @@
...
+- SERVICE_PRINCIPAL_NAME="sp-k2-$unique"
+
+```bash
+vscode ➜ /workspaces/K2Bridge/.devcontainer/adx (feature/script-adx-install ✗) $ bash create-adx.sh -u <uniquename>
+[...]
+Use following settings/secrets in appsettings.development.json:
+aadClientId: 00000000-0000-0000-0000-000000000000
+aadClientSecret: tuUuUuUuU-JXx~x~xxxxxxxxxxxJ00000_
+aadTenantId: 00000000-0000-0000-0000-000000000000
+adxClusterUrl: https://adxk2<uniquename>.westeurope.kusto.windows.net
```
eladiw, line 52: location should not be hard coded (sometimes the location might not be available and you need an easy way to specify another)

eladiw (changes requested):
some comments

.devcontainer/adx/create-adx.sh
dupuyjs: Link to the issue added.

.devcontainer/adx/create-adx.sh
dupuyjs: Fixed.

.devcontainer/adx/create-adx.sh
dupuyjs: Good point. I added the location argument.

.devcontainer/adx/create-adx.sh
dupuyjs: Fixed.

.devcontainer/adx/create-adx.sh
dupuyjs: Yes. SP id was already printed at the end of the script; but I added a reminder message.

At the end of your development, don't forget to delete service principal and its role assignements.
az ad sp delete --id b4edb1f6-e538-4071-93ea-xxxxxxxxxxxx

.devcontainer/adx/create-adx.sh
dupuyjs: I removed the cut commands but I prefer to keep native way to query azure cli outputs (I mean when only one value is needed).  So, when possible I use : 
--query <property> -o tsv

But, unfortunately, when I cannot use this pattern (for instance $SECRETS where I need multiple values); I switch to jq.

.devcontainer/adx/create-adx.sh
dupuyjs: Yes, both options are valid.
1 -> If we need to create new sample tables, it will be easier to update the script.
3 -> It avoids to create a storage just for one file.

At the moment I suggest to keep as is.

docs/devcontainer-adx.md
dupuyjs: Fixed.

docs/devcontainer-adx.md
dupuyjs: Fixed.

AsafMah (approved):

eladiw (approved):
lgtm

PR #71, opened 2021-11-23
Feature pipelines

YannickBrombach:
Removing the namespaces deletion during the cleanup phase

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -351,7 +351,7 @@ stages:
           kubectl run --restart=Never \
             --image=$(ACR_NAME).azurecr.io/k2bridge-test:$SEMANTIC_VERSION \
             --env=K2BRIDGE_URL=http://k2bridge:8080 \
-            --env=ELASTICSEARCH_URL=http://elasticsearchqa-"$KUBERNETES_NAMESPACE":9200 \
+            --env=ELASTICSEARCH_URL=http://elasticsearchqa-master:9200 \
             --env=KUSTO_URI=$KUSTO_URI \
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -371,31 +371,31 @@ stages:
...
-      displayName: Delete AKS namespace and Kusto database
-      condition: and(variables['KUBERNETES_NAMESPACE'], not(eq(variables['KUBERNETES_NAMESPACE'], 'master')))
-      inputs:
-        azureSubscription: $(ACR_PULL_SERVICE_CONNECTION)
-        scriptLocation: inlineScript
-        inlineScript: |
-          set -eux  # fail on error
-          kubectl delete namespace "$KUBERNETES_NAMESPACE"
-          az kusto database delete -g "$(RESOURCE_GROUP)" --cluster-name "$(KUSTO_NAME)" -n "$(KUSTO_DB)" -y
+# - stage: cleanup_integration_tests
```
eladiw, line 374: commenting this code, will cause the integration tests to never be deleted.
We need to add an exclusion condition (e.g. line 391)

eladiw (changes requested):
Added comment

azure-pipelines.yml
YannickBrombach: Fixed by adding a condition instead

eladiw (changes requested):
Search for the work 'master' in this file. I think some (but not all) of the places you find needs to be addressed with vnext (example: line 252)

eladiw (approved):
lgtm. lets try it

AsafMah (approved):

PR #75, opened 2021-12-01
Fix match clause for non-string values

AsafMah:


K2Bridge/JsonConverters/MatchPhraseClauseConverter.cs
```diff
@@ -39,7 +39,7 @@ internal class MatchPhraseClauseConverter : ReadOnlyJsonConverter
                 var obj = new MatchPhraseClause
                 {
                     FieldName = first.Name,
-                    Phrase = (string)((JValue)first.First).Value,
+                    Phrase = ((JValue)first.First).Value?.ToString() ?? string.Empty,
```
eladiw, line 42: can you add unit tests for the two options?

YannickBrombach (approved):
Looking good to me!

eladiw (changes requested):
please add UTs

K2Bridge/ElasticQueryTranslator.cs
```diff
@@ -83,7 +83,7 @@ public QueryData TranslateQuery(string header, string query)
                                 elasticSearchDsl.HighlightText.Add("*", queryStringClause.Phrase);
                                 break;
                             case MatchPhraseClause matchPhraseClause:
-                                elasticSearchDsl.HighlightText.Add(matchPhraseClause.FieldName, matchPhraseClause.Phrase);
+                                elasticSearchDsl.HighlightText.Add(matchPhraseClause.FieldName, matchPhraseClause.Phrase.ToString()); // TODO - fix this when highlighting is fixed
```
eladiw, line 86: what is this todo? I am not aware of any issue with the highlighting

K2Bridge.Tests.UnitTests/Visitors/MatchPhraseVisitorTests.cs
```diff
@@ -19,6 +20,30 @@ public string MatchPhraseVisit_WithValidClause_ReturnsEquals()
...
 
+        [TestCase(ExpectedResult = "MyField == \"3\"")]
+        public string MatchPhraseVisit_WithValidIntClause_ReturnsEquals()
+        {
+            var matchPhraseClause = CreateMatchPhraseClause("MyField", 3);
+
+            return VisitQuery(matchPhraseClause);
+        }
+
+        [TestCase(ExpectedResult = "MyField == \"3.5\"")]
```
eladiw, line 31: is this correct? should 3.5 treated as string?

eladiw (changes requested):
some questions

K2Bridge/ElasticQueryTranslator.cs
AsafMah: Do you know how can I actually get to this line?

When adding a match clause in the UI it's always under elasticSearchDsl.Query.Bool.Filter

K2Bridge.Tests.UnitTests/Visitors/MatchPhraseVisitorTests.cs
AsafMah: It does work, but now I changed it to be more idiomatic

K2Bridge/ElasticQueryTranslator.cs
eladiw: try the filters

eladiw (approved):

PR #77, opened 2021-12-02
Fix ranged queries

AsafMah:
Fixes ranged queries, such as:
```
field[1 TO *]
field[* TO 2007-12-31T23:53:00Z]
field{1 to 3}
```

Due to library parsing support we still don't have:
`field[1 TO 3}`
But this code will work given that.

Still missing more extensive tests

### DIFF
--- a/K2Bridge.Tests.UnitTests/Visitors/RangeClauseVisitorTests.cs
+++ b/K2Bridge.Tests.UnitTests/Visitors/RangeClauseVisitorTests.cs
@@ -13,8 +13,8 @@ namespace UnitTests.K2Bridge.Visitors
     public enum RangeType
     {
         Wildcard,
-        Open,
-        Closed,
+        Exclusive,
+        Inclusive,
     }
 
     [TestFixture]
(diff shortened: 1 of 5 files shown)


K2Bridge.Tests.UnitTests/Visitors/RangeClauseVisitorTests.cs
```diff
@@ -10,102 +10,151 @@ namespace UnitTests.K2Bridge.Visitors
     using global::K2Bridge.Visitors;
     using NUnit.Framework;
 
+    public enum RangeType
+    {
+        Wildcard,
+        Open,
```
eladiw, line 16: nit: inclusive, exclusive?

whats the best terms?

eladiw (approved):
Looks good. 
🚀  Nice tests 🚀

K2Bridge.Tests.UnitTests/Visitors/RangeClauseVisitorTests.cs
AsafMah: The docs use inclusive and exclusive https://www.elastic.co/guide/en/elasticsearch/reference/7.15/query-dsl-query-string-query.html#query-string-syntax
So let's go with that

eladiw (approved):

tomconte (approved):

PR #78, opened 2021-12-06
Support for Min aggregation

Mimetis:

* min aggregation


YannickBrombach (commented):
Looking good to me

eladiw (approved):

K2Bridge/Visitors/MetricAggregationVisitor.cs
```diff
@@ -31,6 +31,15 @@ public void Visit(CardinalityAggregation cardinalityAggregation)
             cardinalityAggregation.KustoQL = $"['{cardinalityAggregation.FieldAlias}']={KustoQLOperators.DCount}({cardinalityAggregation.FieldName})";
         }
 
+        /// <inheritdoc/>
+        public void Visit(MinAggregation minAggregation)
```
AsafMah, line 35: Nit: in all places except here max comes before min. Personally I think min should become before max, but it's up to you, just stay consistent.

AsafMah (approved):
Looks good

K2Bridge/Visitors/MetricAggregationVisitor.cs
Mimetis: Indeed, make more sense. More readable, from a human perspective

PR #79, opened 2021-12-06
feat: Range aggregation

tomconte:
Implementation notes about the Range aggregation. 

* The generated Kusto query uses a `union` operator to aggregate sub-queries, one per range. This is more complex than the logic for the existing bucket aggregations, which only added statements to an ongoing `summarize` operator. In this PR, the responsibility for assembling the full query, including the metrics, is pushed down to the bucket aggregation Visitor. This allows the Range visitor to generate a full query, where the metrics expressions are re-used for each bucket.
* The Range aggregation supports (and Kibana uses) a "keyed" [bucket format](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-range-aggregation.html#_keyed_response_4) that is different from other types of buckets: the values are returned as a key/value hash and not as an array. This is implemented using a new `Keyed` boolean on `BucketsCollection`.


AsafMah:
Haven't got a chance to look though yet, but do note that `RangeClause` supports all of the combinations,
you can see how it's used in my PR - https://github.com/microsoft/K2Bridge/pull/77/files

AsafMah (changes requested):

K2Bridge/Factories/BucketFactory.cs
```diff
@@ -84,5 +85,60 @@ public static TermsBucket CreateTermsBucketFromDataRow(DataRow row)
...
+        /// <summary>
+        /// Create a new <see cref="RangeBucket" from a given <see cref="DataRow"/>/>.
+        /// </summary>
+        /// <param name="row">The row to be transformed to bucket.</param>
+        /// <returns>A new TermsBucket.</returns>
+        public static RangeBucket CreateRangeBucketFromDataRow(DataRow row)
+        {
+            Ensure.IsNotNull(row, nameof(row));
+
+            var range = Convert.ToString(row[(int)BucketColumnNames.Timestamp]);
```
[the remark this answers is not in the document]
tomconte: Actually since the constant is used in several places, I renamed it to `SummarizeByColumn` to reflect that it is really the indx of the column created by `summarize ... by`.

K2Bridge/Factories/BucketFactory.cs
```diff
@@ -84,5 +85,60 @@ public static TermsBucket CreateTermsBucketFromDataRow(DataRow row)
...
+            // Ignore the row for "other" records, that did not match the ranges
+            if (range == BucketColumnNames.RangeDefaultBucket)
+            {
+                return null;
+            }
+
+            // Parse the range
+            string[] splitRange = range.Split('-');
+            double? from = string.IsNullOrEmpty(splitRange[0]) ? null : Convert.ToDouble(splitRange[0]);
+            double? to = string.IsNullOrEmpty(splitRange[1]) ? null : Convert.ToDouble(splitRange[1]);
```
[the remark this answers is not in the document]
tomconte: 👍🏻

K2Bridge/Factories/BucketFactory.cs
```diff
@@ -84,5 +85,60 @@ public static TermsBucket CreateTermsBucketFromDataRow(DataRow row)
...
+
+            // Assemble the key
+            string key = (from, to) switch
+            {
+                (null, null) => "*-*",
+                (_, null) => $"{from}-*",
+                (null, _) => $"*-{to}",
+                (_, _) => $"{from}-{to}",
+            };
```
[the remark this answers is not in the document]
tomconte: 👍🏻

K2Bridge/Factories/BucketFactory.cs
```diff
@@ -84,5 +85,60 @@ public static TermsBucket CreateTermsBucketFromDataRow(DataRow row)
...
+            var rb = new RangeBucket
+            {
+                DocCount = Convert.ToInt32(count),
+                Key = key,
+                From = from,
+                To = to,
+                Aggs = new System.Collections.Generic.Dictionary<string, System.Collections.Generic.List<double>>(),
+            };
+
+            var clmns = row.Table.Columns;
```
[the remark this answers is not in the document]
tomconte: 👍🏻

I also refactored to reuse the code.

K2Bridge/JsonConverters/DateHistogramBucketAggsConverter.cs
```diff
@@ -4,6 +4,7 @@
 
 namespace K2Bridge.JsonConverters
 {
+    using System;
```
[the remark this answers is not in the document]
tomconte: 👍🏻

K2Bridge/JsonConverters/RangeBucketAggsConverter.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// <summary>
+    /// A converter able to serialize Bucket aggregations to Elasticsearh response json format.
+    /// </summary>
+    internal class RangeBucketAggsConverter : WriteOnlyJsonConverter
+    {
+        /// <inheritdoc/>
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            var buck = value as RangeBucket;
```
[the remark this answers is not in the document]
tomconte: Fixed it here and in the other bucket aggs converters.

K2Bridge.Tests.UnitTests/Visitors/RangeAggregationVisitorTests.cs
```diff
@@ -0,0 +1,40 @@
...
+    using System.Collections.Generic;
+    using global::K2Bridge.Models.Request.Aggregations;
+    using global::K2Bridge.Tests.UnitTests.Visitors;
+    using global::K2Bridge.Visitors;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class RangeAggregationVisitorTests
+    {
+        [TestCase(0, 800, 800, 2000, ExpectedResult = "wibble by ['_alias'] = case(dayOfWeek >= 0 and dayOfWeek < 800, '0-800', dayOfWeek >= 800 and dayOfWeek < 2000, '800-2000', 'default_bucket')")]
```
eladiw, line 16: nit: In the latet version alias are no longer prefixed with '_'

K2Bridge/Factories/BucketFactory.cs
```diff
@@ -26,7 +27,7 @@ public static DateHistogramBucket CreateDateHistogramBucketFromDataRow(DataRow r
             // TODO: timestamp is always the first row (probably named _2), and count will be named _count
             // we currently mix index and column names, need to check if we can enhance this logic
             // workitem 15050
-            var timestamp = row[(int)BucketColumnNames.Timestamp];
+            var timestamp = row[(int)BucketColumnNames.SummarizeByColumn];
```
eladiw, line 30: is it still 'timestamp' or maybe
var summarize 
?

K2Bridge/JsonConverters/RangeBucketAggsConverter.cs
```diff
@@ -0,0 +1,68 @@
...
+                serializer.Serialize(writer, buck.To);
+            }
+
+            if (buck.Key is not null)
+            {
+                writer.WritePropertyName("key");
+                serializer.Serialize(writer, buck.Key);
+            }
+
+            foreach (var agg in aggs.Keys)
```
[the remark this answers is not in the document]
tomconte: I have refactored to reuse the code. Nota bene: this section of the code will be reviewed by @dupuyjs for simplification.

K2Bridge/Visitors/KustoQLOperators.cs
```diff
@@ -50,6 +50,8 @@ public static class KustoQLOperators
         public const string MatchRegex = "matches regex";
         public const string GetSchema = "getschema";
         public const string Equal = "==";
+        public const string Extend = "extend";
```
eladiw, line 53: is extend still in use?

K2Bridge/Visitors/RangeVisitor.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            // case() predicates
+            // foo > 0 and foo < 800, "0-800"
+            foreach (var range in rangeAggregation.Ranges)
+            {
+                var rangeClause = new Models.Request.Queries.RangeClause()
+                {
+                    FieldName = rangeAggregation.Field,
+                    GTEValue = range.From.HasValue ? range.From.ToString() : "*",
+                    LTValue = range.To.HasValue ? range.To.ToString() : "*",
```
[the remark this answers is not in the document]
tomconte: 👍🏻

K2Bridge/KustoDAL/KustoResponseParser.cs
```diff
@@ -190,7 +194,10 @@ private void ReportQueryExecutionMetrics(KustoResponseDataSet kustoResponseDataS
                     foreach (DataRow row in parsedKustoResponse[AggregationTableName].TableData.Rows)
                     {
                         IBucket bucket = createBucketFromDataRow(row);
-                        response.AddBucketToAggregation(bucket);
+                        if (bucket != null)
```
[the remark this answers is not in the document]
tomconte: This allows to skip a bucket, for example in the range aggregation one of the rows is dropped (the one for "other" values).

K2Bridge.Tests.UnitTests/KustoDAL/KustoResponseParserTests.cs
```diff
@@ -249,5 +269,36 @@ private static DataTable GetAggsTable()
 
             return resTable;
         }
+
+        private static DataTable GetRangeAggsTable()
+        {
+            DataTable resTable = new DataTable();
+
+            var column1 = new DataColumn("_2");
```
dupuyjs, line 277: As we don't use any longer _ prefix, I suggest to remove also this hack from tests.

K2Bridge.Tests.UnitTests/KustoDAL/KustoResponseParserTests.cs
```diff
@@ -249,5 +269,36 @@ private static DataTable GetAggsTable()
...
+            var column2 = new DataColumn("count_");
+
+            resTable.Columns.Add(column1);
+            resTable.Columns.Add(column2);
+
+            var row1 = resTable.NewRow();
+            row1["_2"] = "-100";
```
dupuyjs, line 284: Same previous comment for all rows.

K2Bridge.Tests.UnitTests/Visitors/RangeAggregationVisitorTests.cs
tomconte: 👍🏻

K2Bridge.Tests.UnitTests/KustoDAL/KustoResponseParserTests.cs
tomconte: 👍🏻

K2Bridge.Tests.UnitTests/KustoDAL/KustoResponseParserTests.cs
tomconte: 👍🏻

K2Bridge/Factories/BucketFactory.cs
tomconte: 👍🏻

K2Bridge/Models/Response/RangeBucket.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// </summary>
+        [JsonProperty("to")]
+        public double? To { get; set; }
+
+        /// <summary>
+        /// Gets or sets the bucket key.
+        /// This key is a string containing the term.
+        /// </summary>
+        [JsonProperty("key")]
+        public string Key { get; set; }
```
eladiw, line 40: you also have this property in 
IKeyedBucket

second question, is IKeyedBucket really used somewhere else? do we prefer to keep it?

K2Bridge/Models/Response/IKeyedBucket.cs
```diff
@@ -0,0 +1,20 @@
...
+// See LICENSE file in the project root for full license information.
+
+namespace K2Bridge.Models.Response
+{
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// An interface for a bucket class.
+    /// </summary>
+    public interface IKeyedBucket : IBucket
```
eladiw, line 12: Is the plan to use this interface on other buckets as well?

K2Bridge/Visitors/RangeVisitor.cs
```diff
@@ -0,0 +1,52 @@
...
+            var caseExpression = $"{KustoQLOperators.Case}(";
+
+            // case() predicates
+            // foo > 0 and foo < 800, "0-800"
+            foreach (var range in rangeAggregation.Ranges)
+            {
+                var rangeClause = new Models.Request.Queries.RangeClause()
+                {
+                    FieldName = rangeAggregation.Field,
+                    GTEValue = range.From?.ToString() ?? "*",
```
eladiw, line 34: in case we have to and from empty
we get
*-* 
?

if so maybe we can skip and not add this to the query (optimization)

K2Bridge/Visitors/RangeVisitor.cs
```diff
@@ -0,0 +1,52 @@
...
+                    LTValue = range.To?.ToString() ?? "*",
+                };
+                rangeClause.Accept(this);
+
+                var bucketName = $"{range.From}-{range.To}";
```
eladiw, line 39: logical question, 
what happens if you have 2 ranges
1-100
50-90
?

eladiw: I guess we should stick to the behavior of Elastic. see what happens.

K2Bridge.Tests.UnitTests/KustoDAL/KustoResponseParserTests.cs
```diff
@@ -156,6 +157,25 @@ public void ParseElasticResponse_WithAggs_ReturnsElasticResponseWithAggs()
             Assert.AreEqual(2, elasticResult.Aggregations.Collection.Buckets.Count());
         }
 
+        [Test]
```
eladiw, line 160: In the visitor file, below, I added a comment about what happens with overlapping ranges
1-100
10-30

Depends on what we decide should be the behavior maybe add a test to validate it

K2Bridge/Factories/BucketFactory.cs
```diff
@@ -23,12 +24,12 @@ public static DateHistogramBucket CreateDateHistogramBucketFromDataRow(DataRow r
         {
             Ensure.IsNotNull(row, nameof(row));
 
-            // TODO: timestamp is always the first row (probably named _2), and count will be named _count
+            // TODO: timestamp is always the first column (probably named 2), and count will be named _count
```
eladiw, line 27:
```suggestion
            // TODO: The 'summary by' field is always the first column and count will be named _count
```

K2Bridge/JsonConverters/RangeBucketAggsConverter.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// <summary>
+    /// A converter able to serialize Bucket aggregations to Elasticsearh response json format.
+    /// </summary>
+    internal class RangeBucketAggsConverter : BucketAggsJsonConverter
+    {
+        /// <inheritdoc/>
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            var buck = (RangeBucket)value;
```
eladiw, line 19: nit:

```suggestion
            var bucket = (RangeBucket)value;
```

K2Bridge/JsonConverters/TermsBucketAggsConverter.cs
```diff
@@ -11,12 +11,12 @@ namespace K2Bridge.JsonConverters
...
     /// A converter able to serialize Bucket aggregations to Elasticsearh response json format.
     /// </summary>
-    internal class TermsBucketAggsConverter : WriteOnlyJsonConverter
+    internal class TermsBucketAggsConverter : BucketAggsJsonConverter
     {
         /// <inheritdoc/>
         public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
         {
-            var buck = value as TermsBucket;
+            var buck = (TermsBucket)value;
```
eladiw, line 19: nit
```suggestion
            var bucket = (TermsBucket)value;
```

eladiw (changes requested):
Some comments

K2Bridge/Models/Response/RangeBucket.cs
tomconte: Since `IBucket` and `IKeyedBucket` only contained properties, I refactored to use a base class instead... Let me know what you think.

K2Bridge/Models/Response/IKeyedBucket.cs
tomconte: Refactored to use a base class.

K2Bridge/Visitors/RangeVisitor.cs
tomconte: Actually this is a valid case, minus infinity to plus infinity, but this was not properly handled in the Kusto query... It is fixed (+ corresponding test).

K2Bridge/Factories/BucketFactory.cs
tomconte: I have reverted to `timestamp`.

K2Bridge/JsonConverters/TermsBucketAggsConverter.cs
tomconte: 👍🏻

K2Bridge/Visitors/RangeVisitor.cs
tomconte: After testing these cases (overlapping range), it turns out that the `case()` based Kusto query does not cover these. I am working on implementing the `union()` query that Asaf mentioned (see update in the Wiki page) at the beginning and it is working for these cases.

Because the query is more complex, it requires a bunch of changes to the Visitors (but Request/Response are unchanged).

Would you like me to add more commits to this already large PR, or should we merge it and then I can open another one today with the query changes. LMKWYT! @AsafMah @eladiw

K2Bridge.Tests.UnitTests/KustoDAL/KustoResponseParserTests.cs
tomconte: The new query should return the same results as the previous one, so there should be no changes to the response parser.

K2Bridge/Visitors/RangeVisitor.cs
tomconte: By popular vote, I will be pushing more commits :-)

K2Bridge.Tests.UnitTests/Visitors/AggregationVisitorTests.cs
```diff
@@ -37,24 +37,6 @@ public string AggregationVisit_WithAvgAgg_ReturnsAvgPrimary()
             return aggregateClause.KustoQL;
         }
 
-        [TestCase(ExpectedResult = "['B']=avg(fieldB), ['A']=avg(fieldA)")]
```
tomconte, line 40: Nota bene: I removed the test cases where a metric aggregation had sub aggregations; this does not happen in practice:

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket.html

> Bucket aggregations, **as opposed to metrics aggregations**, can hold sub-aggregations.

K2Bridge.Tests.End2End/QueryControllerErrorsTests.cs
```diff
@@ -38,6 +37,7 @@ public class QueryControllerErrorsTests : KustoTestBase
         public async Task MSearch_WithInvalidQuery_ReturnsElasticErrorResponse()
         {
             var response = await K2Client().MSearch(INDEX, $"{FLIGHTSDIR}/MSearch_Invalid_Query.json", false);
+            response.SelectToken("responses[*].error.root_cause[*].reason").Parent.Remove();
```
eladiw, line 40: why did you remove it

K2Bridge/KustoDAL/KustoResponseParser.cs
```diff
@@ -189,10 +194,20 @@ private void ReportQueryExecutionMetrics(KustoResponseDataSet kustoResponseDataS
...
+                        Bucket bucket = createBucketFromDataRow(row);
+                        if (bucket != null)
+                        {
+                            response.AddBucketToAggregation(bucket);
+                        }
                     }
                 }
+
+                // For Range aggregations, the calculated total hits is wrong, so we have an additional column with the expected count
+                if (parsedKustoResponse[HitsTotalTableName] != null)
```
eladiw, line 206: so this exists only for range query?

eladiw (approved):
looks good.

K2Bridge/KustoDAL/KustoResponseParser.cs
tomconte: Currently, for any aggregation... As an optimization we could emit the `hitsTotal` query for Range only, yes.

https://github.com/microsoft/K2Bridge/pull/79/files/70e32a243b509ee13d457f6beca27f39c08e3f6d#diff-6ee10d9b710b3a5b20475e9968163ec8723bd8e6319f9aed7ed2b23f80695767R72

K2Bridge.Tests.End2End/QueryControllerErrorsTests.cs
tomconte: The `reason` error message contains the query, so every time we change the query the test fails.

K2Bridge/Factories/BucketFactory.cs
```diff
@@ -38,51 +39,104 @@ public static DateHistogramBucket CreateDateHistogramBucketFromDataRow(DataRow r
...
 
-            var tb = new TermsBucket
+            // Ignore the row for "other" records, that did not match the ranges
+            if (range == BucketColumnNames.RangeDefaultBucket)
+            {
+                return null;
+            }
+
+            // Parse the range
+            string[] splitRange = range.Split('-');
```
AsafMah, line 90: I would do it like this:
```c#
var splitRange = range
                .Split('-')
                .Select(s => string.IsNullOrEmpty(s) ? (double?)null : double.Parse(s))
                .ToArray();
var from = splitRange[0];
var to = splitRange[1];
```
But if you prefer yours it's also fine

K2Bridge/KustoDAL/KustoResponseParser.cs
```diff
@@ -173,9 +174,13 @@ private void ReportQueryExecutionMetrics(KustoResponseDataSet kustoResponseDataS
                 response.AddParentToAgg(parent);
 
                 // Determine how to create the buckets based on the aggregation type
-                Func<DataRow, IBucket> createBucketFromDataRow = null;
+                Func<DataRow, Bucket> createBucketFromDataRow = null;
                 switch (query.PrimaryAggregation)
```
AsafMah, line 178: You can simplify to 
```c#
                Func<DataRow, Bucket> createBucketFromDataRow = null;
                switch (query.PrimaryAggregation)
                {
                    case nameof(Models.Request.Aggregations.RangeAggregation):
                        createBucketFromDataRow = BucketFactory.CreateRangeBucketFromDataRow;
                        response.SetAggregationKeyed(true);
                        break;
                    case nameof(Models.Request.Aggregations.TermsAggregation):
                        createBucketFromDataRow = BucketFactory.CreateTermsBucketFromDataRow;
                        break;
                    case nameof(Models.Request.Aggregations.DateHistogramAggregation):
                        createBucketFromDataRow = BucketFactory.CreateDateHistogramBucketFromDataRow;
                        break;
                }
```
And if you prefer:
```c# 
                // Determine how to create the buckets based on the aggregation type
                Func<DataRow, Bucket> createBucketFromDataRow = query.PrimaryAggregation switch {
                    nameof(Models.Request.Aggregations.RangeAggregation) => BucketFactory.CreateRangeBucketFromDataRow,
                    nameof(Models.Request.Aggregations.TermsAggregation) => BucketFactory.CreateTermsBucketFromDataRow,
                    nameof(Models.Request.Aggregations.DateHistogramAggregation) => BucketFactory.CreateDateHistogramBucketFromDataRow,
                    _ => null,
                };
                if (query.PrimaryAggregation == nameof(Models.Request.Aggregations.RangeAggregation))
                {
                    response.SetAggregationKeyed(true);
                }
```

K2Bridge/KustoDAL/KustoResponseParser.cs
```diff
@@ -173,9 +174,13 @@ private void ReportQueryExecutionMetrics(KustoResponseDataSet kustoResponseDataS
                 response.AddParentToAgg(parent);
 
                 // Determine how to create the buckets based on the aggregation type
-                Func<DataRow, IBucket> createBucketFromDataRow = null;
+                Func<DataRow, Bucket> createBucketFromDataRow = null;
```
AsafMah, line 177: Do you know why we used IBucket in the first place? Now that you changed it, it doesn't seem to be used anywhere. So either delete it or find why we need it

K2Bridge/KustoDAL/KustoResponseParser.cs
```diff
@@ -189,10 +194,20 @@ private void ReportQueryExecutionMetrics(KustoResponseDataSet kustoResponseDataS
...
-                        response.AddBucketToAggregation(bucket);
+                        Bucket bucket = createBucketFromDataRow(row);
+                        if (bucket != null)
+                        {
+                            response.AddBucketToAggregation(bucket);
+                        }
                     }
                 }
+
+                // For Range aggregations, the calculated total hits is wrong, so we have an additional column with the expected count
```
AsafMah, line 205: Can you explain further? why is it wrong?

K2Bridge/Visitors/AggregationVisitor.cs
```diff
@@ -21,19 +21,28 @@ public void Visit(AggregationContainer aggregationContainer)
...
+                        // Metrics aggregations just return their KustoQL
+                        aggregationContainer.KustoQL += $"{subAgg.KustoQL}, "; // this won't work when 2+ bucket aggregations are used!
+                    }
                 }
             }
 
-            aggregationContainer.KustoQL += aggregationContainer.PrimaryAggregation.KustoQL;
+            aggregationContainer.PrimaryAggregation.Accept(this);
+
+            aggregationContainer.KustoQL = aggregationContainer.PrimaryAggregation.KustoQL;
```
AsafMah, line 45: You overwrite here Everything you added in `aggregationContainer.KustoQL += $"{subAgg.KustoQL}, "`

K2Bridge/Visitors/AggregationVisitor.cs
```diff
@@ -21,19 +21,28 @@ public void Visit(AggregationContainer aggregationContainer)
...
+                    else
+                    {
+                        // Metrics aggregations just return their KustoQL
+                        aggregationContainer.KustoQL += $"{subAgg.KustoQL}, "; // this won't work when 2+ bucket aggregations are used!
+                    }
                 }
             }
 
-            aggregationContainer.KustoQL += aggregationContainer.PrimaryAggregation.KustoQL;
+            aggregationContainer.PrimaryAggregation.Accept(this);
```
AsafMah, line 43: Why did you change the order?

K2Bridge/Visitors/AggregationVisitor.cs
```diff
@@ -21,19 +21,28 @@ public void Visit(AggregationContainer aggregationContainer)
...
-            aggregationContainer.PrimaryAggregation.Accept(this);
-
             // TODO: do something with the sub aggregations to KQL
             if (aggregationContainer.SubAggregations != null)
             {
                 foreach (var (_, subAgg) in aggregationContainer.SubAggregations)
                 {
                     subAgg.Accept(this);
-                    aggregationContainer.KustoQL += $"{subAgg.KustoQL}, "; // this won't work when 2+ bucket aggregations are used!
+                    if (aggregationContainer.PrimaryAggregation.GetType().BaseType.Name == "BucketAggregation")
```
AsafMah, line 30: Why are you using reflection here??
```c#
if (aggregationContainer.PrimaryAggregation is BucketAggregation bucketAggregation)
                    {
                        // Bucket aggregations are responsible for inserting the metrics expressions in the final query
                        bucketAggregation.MetricsKustoQL += $"{subAgg.KustoQL}, ";
                    }
```

K2Bridge/Visitors/RangeVisitor.cs
```diff
@@ -0,0 +1,66 @@
...
+                rangeClause.Accept(this);
+
+                if (string.IsNullOrEmpty(rangeClause.KustoQL))
+                {
+                    // This is then "open" range, -infinity to +infinity
+                    rangeAggregation.KustoQL += "true";
+                }
+                else
+                {
+                    rangeAggregation.KustoQL += $"{rangeClause.KustoQL}";
```
AsafMah, line 47: Is this not equivalent to `rangeAggregation.KustoQL += rangeClause.KustoQL;` ?

K2Bridge/Visitors/RangeVisitor.cs
```diff
@@ -0,0 +1,66 @@
...
+                // Followed by the default aggregation metric
+                rangeAggregation.KustoQL += $" | {KustoQLOperators.Summarize} {rangeAggregation.MetricsKustoQL}{rangeAggregation.Metric}";
+
+                // Create the column with the range name
+                var bucketName = $"{range.From}-{range.To}";
+                rangeAggregation.KustoQL += $" | {KustoQLOperators.Extend} ['{rangeAggregation.Key}'] = '{bucketName}'), ";
+            }
+
+            // Remove final comma and space
+            rangeAggregation.KustoQL = rangeAggregation.KustoQL.Substring(0, rangeAggregation.KustoQL.Length - 2);
```
AsafMah, line 60: `rangeAggregation.KustoQL = rangeAggregation.KustoQL.TrimEnd(' ', ',');`

AsafMah (changes requested):

K2Bridge/Visitors/RangeVisitor.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    /// <content>
+    /// A visitor for the <see cref="RangeAggregation"/> element.
+    /// </content>
+    internal partial class ElasticSearchDSLVisitor : IVisitor
+    {
+        /// <inheritdoc/>
+        public void Visit(RangeAggregation rangeAggregation)
+        {
+            Ensure.IsNotNull(rangeAggregation, nameof(TermsAggregation));
```
eladiw, line 19:
```suggestion
            Ensure.IsNotNull(rangeAggregation, nameof(RangeAggregation));
```

K2Bridge.Tests.End2End/QueryControllerErrorsTests.cs
eladiw: The question is, how many times is it changed, the answer is while developing.
the pros, of leaving it, is that its easier to read the code and test.
I vote keep it

K2Bridge.Tests.UnitTests/Visitors/RangeAggregationVisitorTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    using System.Collections.Generic;
+    using global::K2Bridge.Models.Request.Aggregations;
+    using global::K2Bridge.Tests.UnitTests.Visitors;
+    using global::K2Bridge.Visitors;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class RangeAggregationVisitorTests
+    {
+        [TestCase(0, 800, 800, 2000, ExpectedResult = "union (_data | where dayOfWeek >= 0 and dayOfWeek < 800 | summarize wibble | extend ['key'] = '0-800'), (_data | where dayOfWeek >= 800 and dayOfWeek < 2000 | summarize wibble | extend ['key'] = '800-2000') | project-reorder * asc")]
```
eladiw, line 16: about this part:
"project-reorder * asc"

There is a place in the code where we say that we refer to column names both as index and also as name. (and we say we need to fix this)
Wouldn't re ordering might cause issues if referring to a column by index?

K2Bridge/Factories/BucketFactory.cs
eladiw: here, this is the place where the project-reorder might cause issues

K2Bridge/Visitors/RangeVisitor.cs
```diff
@@ -0,0 +1,66 @@
...
+                    rangeAggregation.KustoQL += "true";
+                }
+                else
+                {
+                    rangeAggregation.KustoQL += $"{rangeClause.KustoQL}";
+                }
+
+                // Insert the KustoQL generated by the metrics aggregations
+                // Followed by the default aggregation metric
+                rangeAggregation.KustoQL += $" | {KustoQLOperators.Summarize} {rangeAggregation.MetricsKustoQL}{rangeAggregation.Metric}";
```
eladiw, line 52: so here, you have the MetricsKustoQL which was set up in AggregationVisitor L33, and you use it.
This line is not easy to read. 
a. not sure what the last part means rangeAggregation.Metric
b. the name MetricsKustoQL is not very self explanatory.
c. I would add a lot more comments on this specific flow

eladiw (changes requested):
Some comments

K2Bridge.Tests.End2End/QueryControllerErrorsTests.cs
tomconte: Reverted the change

K2Bridge/Factories/BucketFactory.cs
```diff
@@ -38,51 +39,104 @@ public static DateHistogramBucket CreateDateHistogramBucketFromDataRow(DataRow r
...
+            double? from = string.IsNullOrEmpty(splitRange[0]) ? null : double.Parse(splitRange[0]);
+            double? to = string.IsNullOrEmpty(splitRange[1]) ? null : double.Parse(splitRange[1]);
+
+            // Assemble the key
+            // An empty limit becomes "*"
+            // An integer limit is suffixed with ".0"
+            string fromKey, toKey;
+
+            fromKey = from?.ToString() switch
```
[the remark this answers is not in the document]
tomconte: 👍🏻

K2Bridge/Factories/BucketFactory.cs
tomconte: I switched to your version

K2Bridge/KustoDAL/KustoResponseParser.cs
tomconte: Thanks! I kept the case version but with the simplified syntax.

K2Bridge/KustoDAL/KustoResponseParser.cs
tomconte: `IBucket` was an interface with a single property in it. After the first review pass it did not seem useful to have an interface for this and I switched to a base class. But I forgot to delete the interface source files, this is done now.

K2Bridge/KustoDAL/KustoResponseParser.cs
tomconte: In general, for aggregations, we determine the total hit count by adding the buckets doc counts together. However, this does not work for range buckets *with overlapping ranges*: in this case our total hit count is wrong (or at least, different from what ES returns) because we then count docs twice. ES returns the total hit count over the query but without double counting. I added this just so that we can get the same total hit count as ES.

K2Bridge/Visitors/AggregationVisitor.cs
tomconte: Much better, thanks

K2Bridge/Visitors/AggregationVisitor.cs
tomconte: OK, that else clause here was not needed, I removed it.

K2Bridge/Visitors/AggregationVisitor.cs
tomconte: The Range visitor needs to have all the Metrics KustoQL strings, so that it can use them to assemble its query. So, we first go through all the sub-aggregations (metrics) to gather their KustoQL, and then call the primary aggregation visitor (the bucket aggregation) so that it produces the final query, including the metrics. I have added more comments to explain the flow in this Visitor.

K2Bridge/Visitors/RangeVisitor.cs
tomconte: Absolutely; copy/paste gone wrong :) Fixed.

K2Bridge/Visitors/RangeVisitor.cs
tomconte: 👍🏻

K2Bridge/Visitors/RangeVisitor.cs
tomconte: 👍🏻

K2Bridge.Tests.UnitTests/Visitors/RangeAggregationVisitorTests.cs
tomconte: Just for the Range aggregation, I need this re-order for our column indices to work. I agree we should probably switch to indexing the columns by name instead of indices, it would be more robust.

K2Bridge/Visitors/RangeVisitor.cs
tomconte: I have added more comments here and in `AggregationVisitor` also. Hope it makes sense.

nebrass (approved):

dupuyjs (approved):
lgfm

K2Bridge/Factories/BucketFactory.cs
AsafMah: Why was it resolved? it's still `var timestamp`

tomconte: We discussed offline with Elad; in the case of a DateHistogramBucket, the "summarize by" column is a timestamp indeed, so the name of the variable is appropriate.

AsafMah (approved):

K2Bridge.Tests.UnitTests/Visitors/RangeAggregationVisitorTests.cs
eladiw: what happens if you dont re-order?

K2Bridge/Visitors/RangeVisitor.cs
```diff
@@ -0,0 +1,77 @@
...
+            foreach (var range in rangeAggregation.Ranges)
+            {
+                rangeAggregation.KustoQL += $"(_data | {KustoQLOperators.Where} ";
+
+                var rangeClause = new Models.Request.Queries.RangeClause()
+                {
+                    FieldName = rangeAggregation.Field,
+                    GTEValue = range.From?.ToString() ?? "*",
+                    LTValue = range.To?.ToString() ?? "*",
+                };
```
eladiw, line 45: I think this section should happen at the json to models conversion part.

because, line 40, where you create the clause is not the natural place to create a new clause object. wdyt?

K2Bridge/Visitors/RangeVisitor.cs
eladiw: I know it works, but I am not sure why. 

shouldnt it be:

rangeAggregation.KustoQL += $" | {KustoQLOperators.Summarize} {rangeClause.MetricsKustoQL}{rangeAggregation.Metric}";

notice the 'rangeClause'

K2Bridge/Visitors/RangeVisitor.cs
```diff
@@ -0,0 +1,77 @@
...
+        public void Visit(RangeAggregation rangeAggregation)
+        {
+            Ensure.IsNotNull(rangeAggregation, nameof(RangeAggregation));
+            EnsureClause.StringIsNotNullOrEmpty(rangeAggregation.Metric, nameof(RangeAggregation.Metric));
+            EnsureClause.StringIsNotNullOrEmpty(rangeAggregation.Field, nameof(RangeAggregation.Field));
+
+            // Start the union operator
+            rangeAggregation.KustoQL += "union ";
+
+            // Query expressions for each range
```
eladiw, line 26: great comment

K2Bridge/Visitors/RangeVisitor.cs
tomconte: Well the idea was to reuse the RangeClause to assemble part of the query, here in the RangeVisitor, since it happens to produce the appropriate KustoQL... I guess we could assemble the RangeClause at conversion time, but right now there is no custom JSON converter for the RangeAggregation, since we are trying to simplify that step. The model is straight generated from incoming JSON.

eladiw: I understand, but still. we can only simplify so much. and it seems we moved code of the convertor to the visitor just because we want to keep the convertor simple (or non existent)..

K2Bridge/Visitors/RangeVisitor.cs
tomconte: The `rangeClause` is inserted in the Kusto query in line 56. In line 63, we insert the summarize, then the list of metrics (avg, sum) and finally the mandatory count().

K2Bridge.Tests.UnitTests/Visitors/RangeAggregationVisitorTests.cs
tomconte: Because of the `extend` statement, the column with the range ends up last, instead of first. So, I re-order to put it first, which is what the response parser expects.

K2Bridge/Visitors/RangeVisitor.cs
tomconte: OK, I guess this is a design decision we should discuss with the team to make sure we are aligned. Indeed, I think we have been moving some logic out of the converters, to centralize it in the visitors.

eladiw: lets discuss today

dupuyjs: For me, the serialization is good. We populate the RangeAggregationExpression collection with the correct fields. Then we have a decision later in the visitor 1/ Should we convert RangeAggregationExpression to RangeClause to reuse the visitor associated with -or 2/ Should we create a dedicated visitor for  RangeAggregationExpression.

tomconte: I think Elad's point is: the Model should not necessarily represent the JSON, it should represent whatever the Visitor needs, and it is the conversion layer's job to assemble this Model. So, for example, here, we could have a list of RangeClause in the model instead of RangeAggregationExpression, and a custom converter to create it. In this way the Visitor does not have to care about conversion.

tomconte: As discussed, just pushed the `RangeAggregationExpression` visitor.

K2Bridge/Visitors/RangeVisitor.cs
tomconte: I have renamed `MetricsKustoQL` to `SubAggregationsKustoQL`. Hope it makes more sense.

AsafMah (approved):

K2Bridge/Visitors/RangeVisitor.cs
```diff
@@ -0,0 +1,72 @@
...
+                    // This is then "open" range, -infinity to +infinity
+                    rangeAggregation.KustoQL += "true";
+                }
+                else
+                {
+                    rangeAggregation.KustoQL += range.KustoQL;
+                }
+
+                // Insert the KustoQL generated by the metrics aggregations
+                // rangeAggregation.MetricsKustoQL contains the KustoQL generated by the metrics sub-aggregations (if any)
```
eladiw, line 55:
```suggestion
                // rangeAggregation.SubAggregationsKustoQL contains the KustoQL generated by the metrics sub-aggregations (if any)
```

K2Bridge/Visitors/RangeVisitor.cs
```diff
@@ -0,0 +1,72 @@
...
+                rangeAggregation.KustoQL += $"(_data | {KustoQLOperators.Where} ";
+
+                range.Field = rangeAggregation.Field;
+                range.Accept(this);
+
+                // Insert the range clause
+                if (string.IsNullOrEmpty(range.KustoQL))
+                {
+                    // This is then "open" range, -infinity to +infinity
+                    rangeAggregation.KustoQL += "true";
```
eladiw, line 47: really needed? why not 'continue'?

eladiw (approved):
approved with minor comments

K2Bridge/Visitors/RangeVisitor.cs
tomconte: 👍🏻

K2Bridge/Visitors/RangeVisitor.cs
tomconte: Good catch, thanks! The if statement can be simplified and the query too.

PR #80, opened 2021-12-06
update: Aggregations serialization (request)

dupuyjs:
The following changes are proposed : 

- Remove most of json converters. Only two are remaining:
   - AggregationContainerConverter
   - TermsOrderConverter (used by TermsAggregation)

- Update some naming convention: 
  - Aggregation class is the base object for MetricAggregation and BucketAggregation. (instead of LeafAggregation)
  - AggregationContainer contains PrimaryAggregation and SubAggregations (instead of Aggregation)
  - Key replace FieldAlias
  - Field replace FieldName (to be consistent with other property names)

- Fix potential issue where aggs and primary aggregation could trigger a serialization issue if elements were inversed in json payload.

- BucketAggregation Metric is now a readonly property.

- Remove usage of _ to prefix column names (quotes are used instead).


K2Bridge/JsonConverters/AggregationContainerConverter.cs
```diff
@@ -0,0 +1,80 @@
...
+        public override object ReadJson(
+            JsonReader reader,
+            Type objectType,
+            object existingValue,
+            JsonSerializer serializer)
+        {
+            var jo = JObject.Load(reader);
+
+            var obj = new AggregationContainer();
+            var primary = jo.Children<JProperty>().Where(x => x.Name != "aggs").FirstOrDefault();
```
AsafMah, line 29: Can simplify to
```suggestion
            var primary = jo.Children<JProperty>().FirstOrDefault(x => x.Name != "aggs");
```

K2Bridge/JsonConverters/AggregationContainerConverter.cs
```diff
@@ -0,0 +1,80 @@
...
+                        aggregation = property.Value.ToObject<CardinalityAggregation>(serializer);
+                        break;
+
+                    case "max":
+                        aggregation = property.Value.ToObject<MaxAggregation>(serializer);
+                        break;
+
+                    case "sum":
+                        aggregation = property.Value.ToObject<SumAggregation>(serializer);
+                        break;
```
AsafMah, line 67: Can simplify to
```suggestion
                aggregation = property.Name switch
                {
                    "date_histogram" => property.Value.ToObject<DateHistogramAggregation>(serializer),
                    "terms" => property.Value.ToObject<TermsAggregation>(serializer),
                    "avg" => property.Value.ToObject<AverageAggregation>(serializer),
                    "cardinality" => property.Value.ToObject<CardinalityAggregation>(serializer),
                    "max" => property.Value.ToObject<MaxAggregation>(serializer),
                    "sum" => property.Value.ToObject<SumAggregation>(serializer),
                    _ => aggregation
                };
```

AsafMah (commented):
Overall looks good, don't really have anything except minor suggestions.

I would note that this PR removes some of the logic from the converters to the visitors, which I assume it what we want?

I'll approve when the CI will run again

K2Bridge.Tests.UnitTests/JsonConverters/AggregationConverterTests.cs
```diff
@@ -73,143 +73,150 @@ public class AggregationConverterTests
...
-                        PrimaryAggregation = new TermsAggregation {
-                            FieldName = "DestCountry",
-                            FieldAlias = "_2",
-                            Metric = "count()",
-                            SortFieldName = "_count",
-                            SortOrder = "desc",
+                    PrimaryAggregation = new TermsAggregation
+                    {
+                        Field = "DestCountry",
+                        Key = "2",
```
eladiw, line 114: IMO, alias was more self explanatory

K2Bridge.Tests.UnitTests/JsonConverters/AggregationConverterTests.cs
```diff
@@ -73,143 +73,150 @@ public class AggregationConverterTests
...
+                        Order = new TermsOrder {
+                            Field = "_count",
+                            Sort = "desc",
```
eladiw, line 117: IMO, sortOrder was more self explanatory

K2Bridge.Tests.UnitTests/JsonConverters/AggregationConverterTests.cs
```diff
@@ -73,143 +73,150 @@ public class AggregationConverterTests
...
-                    { "2", new Aggregation() {
-                        PrimaryAggregation = new AvgAggregation {
-                            FieldAlias = "_2",
-                            FieldName = "metric",
-                            },
-                        SubAggregations = new Dictionary<string, Aggregation>(),
-                        }
+                    PrimaryAggregation = new AverageAggregation
+                    {
+                        Field = null,
```
eladiw, line 151: metric or null?

K2Bridge.Tests.UnitTests/JsonConverters/AggregationConverterTests.cs
```diff
@@ -73,143 +73,150 @@ public class AggregationConverterTests
...
-                    { "2", new Aggregation() {
-                        PrimaryAggregation = new AvgAggregation {
-                            FieldAlias = "_2",
-                            FieldName = null,
-                            },
-                        SubAggregations = new Dictionary<string, Aggregation>(),
-                        }
+                    PrimaryAggregation = new CardinalityAggregation
+                    {
+                        Field = "metric",
```
eladiw, line 168: should be null?

eladiw: ok now I see that the order of tests were changed. can you change the order to reduce diff changes

K2Bridge.Tests.UnitTests/Visitors/DateHistogramAggregationVisitorTests.cs
```diff
@@ -11,14 +11,13 @@ namespace UnitTests.K2Bridge.Visitors
     [TestFixture]
     public class DateHistogramAggregationVisitorTests
     {
-        [TestCase(ExpectedResult = "wibble by _alias = wobble\n| order by _alias asc")]
+        [TestCase(ExpectedResult = "count() by ['key'] = field\n| order by ['key'] asc")]
```
eladiw, line 14: so now count() is default?
you removed the metric name

K2Bridge/Factories/BucketFactory.cs
```diff
@@ -46,7 +46,7 @@ public static DateHistogramBucket CreateDateHistogramBucketFromDataRow(DataRow r
                     continue;
                 }
 
-                dhb.Aggs[clmn.ColumnName.Substring(1)] = new System.Collections.Generic.List<double>() { Convert.ToDouble(row[clmn.ColumnName]) };
+                dhb.Aggs[clmn.ColumnName] = new System.Collections.Generic.List<double>() { Convert.ToDouble(row[clmn.ColumnName]) };
```
eladiw, line 49: 👍

K2Bridge/Models/Request/Aggregations/Aggregation.cs
```diff
@@ -4,32 +4,19 @@
...
 {
-    using System.Collections.Generic;
     using K2Bridge.JsonConverters;
     using K2Bridge.Models.Request;
     using K2Bridge.Visitors;
     using Newtonsoft.Json;
 
     /// <summary>
-    /// Aggregation element in an Elasticsearch query.
+    /// Descibes base aggregation class to visit.
```
eladiw, line 13:
```suggestion
    /// Describes base aggregation class to visit.
```

K2Bridge/Models/Request/Aggregations/Metric/Max/MaxAggregation.cs
```diff
@@ -11,7 +11,6 @@ namespace K2Bridge.Models.Request.Aggregations
     /// <summary>
     /// A single-value metrics aggregation that computes the max of numeric values that are extracted from the aggregated documents.
```
eladiw, line 12: Is the 'max' directory needed? same for the rest (avg etc)

K2Bridge/JsonConverters/AggregationContainerConverter.cs
dupuyjs: Awesome. Thanks.

eladiw (changes requested):
Some comments. looks really good

K2Bridge/JsonConverters/AggregationContainerConverter.cs
dupuyjs: Fixed.

K2Bridge/JsonConverters/AggregationContainerConverter.cs
dupuyjs: Fixed.

K2Bridge.Tests.UnitTests/Visitors/DateHistogramAggregationVisitorTests.cs
dupuyjs: Yes. In my opinion, Metric will be always count() for all Bucket aggregation and this metric will never change. If you request another default metric (for instance Average), a sub aggregation is requested, it does not change the default count() metric.

So I updated the property to readonly. Anyway I can rollback if you consider there is a chance it will not be the case. @tomconte

K2Bridge/Factories/BucketFactory.cs
dupuyjs: Thanks.

K2Bridge/Models/Request/Aggregations/Aggregation.cs
dupuyjs: Fixed.

K2Bridge.Tests.UnitTests/JsonConverters/AggregationConverterTests.cs
dupuyjs: Yes. I was hesitating with this change. When reading the code, I realized this property indicates the associated dictionary key where aggregation is located. So I updated the name with this logic. But Alias is also fine for me and I can rollback this change.  @tomconte @nebrass -> Your opinion ?

AsafMah: Personally I think Key is more relevant than alias

K2Bridge/Models/Request/Aggregations/Metric/Max/MaxAggregation.cs
dupuyjs: I need to admit this is a personal preference to access code quickly. And will structure better the code if a Bucket/Metric need several files (at the moment this is only the case for Terms). @nebrass @tomconte -> Your opinion ?

K2Bridge.Tests.UnitTests/JsonConverters/AggregationConverterTests.cs
tomconte: +1 for Key, since this represents the key that identifies the aggregation; the value is used as a field alias in the KustoQL query, but I think from the request/response model perspective, Key makes sense.

K2Bridge.Tests.UnitTests/Visitors/DateHistogramAggregationVisitorTests.cs
tomconte: As far as I can tell, ES bucket aggregations always include a count, so the count() will always be there and will always be a count. That said since the Metric property still exists, I would use it in the tests anyway instead of hard coding the value...

K2Bridge.Tests.UnitTests/JsonConverters/AggregationConverterTests.cs
dupuyjs: Yes. Good point. Order has been changed to default one.

K2Bridge.Tests.UnitTests/JsonConverters/AggregationConverterTests.cs
dupuyjs: Tests order has been changed to default one.

K2Bridge/Models/Request/Aggregations/Metric/Max/MaxAggregation.cs
tomconte: Looks good to me, but then we should probably use the same kind of structure for the other parts of the code.

K2Bridge.Tests.UnitTests/Visitors/DateHistogramAggregationVisitorTests.cs
YannickBrombach: I think this makes sense, simplifies readability, and can go to something more generic if the need arises, which is not the case today

YannickBrombach (commented):
Looking good to me, great changes :)

K2Bridge.Tests.UnitTests/JsonConverters/AggregationConverterTests.cs
dupuyjs: I agree. Fixed the code to use SortField and SortOrder.

K2Bridge.Tests.UnitTests/JsonConverters/AggregationConverterTests.cs
dupuyjs: Key is approved 🥇

K2Bridge.Tests.UnitTests/Visitors/DateHistogramAggregationVisitorTests.cs
dupuyjs: I discussed the point with @tomconte, and I will come back to Metric with a get/set. We can postpone this discussion later, when we'll cover all the aggregations buckets to decide if we can pass the property to readonly.

K2Bridge/Models/Request/Aggregations/Metric/Max/MaxAggregation.cs
dupuyjs: Yes, that's the plan for the answer.

K2Bridge/Models/Request/Aggregations/AggregationDictionary.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT license.
+// See LICENSE file in the project root for full license information.
+
+namespace K2Bridge.Models.Request.Aggregations
+{
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Diagnostics.CodeAnalysis;
+
+    internal class AggregationDictionary : Dictionary<string, AggregationContainer>
```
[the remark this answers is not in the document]
dupuyjs: Fixed.

K2Bridge/Models/Request/Aggregations/AggregationDictionary.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// See LICENSE file in the project root for full license information.
+
+namespace K2Bridge.Models.Request.Aggregations
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Decribes aggregation dictionary element in an Elasticsearch query.
```
eladiw, line 10:
```suggestion
    /// Describes aggregation dictionary element in an Elasticsearch query.
```

K2Bridge/Models/Request/Aggregations/Metric/Max/MaxAggregation.cs
eladiw: so do we add a user story to align the rest of the code?

K2Bridge/Models/Request/Aggregations/AggregationDictionary.cs
dupuyjs: Good point. Thanks.

dupuyjs: Fixed.

eladiw (approved):
lgtm

PR #82, opened 2021-12-07
Fix highlight in filters for kibana 7

AsafMah:
In kibana 7, the filters that are highlighted are under ElasticSearchDsl.Query.Bool.Filter,
So this PR checks them.

Where should the tests for it go? I can't find the highlight tests (if there are any)


K2Bridge/ElasticQueryTranslator.cs
```diff
@@ -67,23 +67,17 @@ public QueryData TranslateQuery(string header, string query)
                 if (elasticSearchDsl.Query.Bool != null)
                 {
                     Ensure.IsNotNull(elasticSearchDsl.Query.Bool.Must, nameof(elasticSearchDsl.Query.Bool.Must));
+                    Ensure.IsNotNull(elasticSearchDsl.Query.Bool.Filter, nameof(elasticSearchDsl.Query.Bool.Filter));
 
-                    if (elasticSearchDsl.Query.Bool.Filter.Any())
```
eladiw, line 71: I think its better to keep the 'using kql' warning. wdyt?

tamirkamara: Filters are not longer an indication for kql syntax as they used even in regular cases (like the date range, etc.)
Perhaps there's something else that could be used...

eladiw (approved):

tamirkamara (approved):

PR #88, opened 2021-12-15
Update the node pools vms sku

eladiw:

* Update the node pools vms sku


tamirkamara (approved):

YannickBrombach (approved):
lgtm

PR #89, opened 2021-12-15
Update the node pool VMs sku (#88)

eladiw:
(cherry picked from commit fdcef0febf92a2bde7559134bca26e613fb655dc)

* Update the node pool VMs sku


AsafMah:
Is there a specific reason?

eladiw:
> Is there a specific reason?

updated aks + to a newer sku

AsafMah (approved):

PR #90, opened 2021-12-15
Code Design Discussion

dupuyjs:

* Update access modifiers to pass aggregation dictionary in QueryData

Context: 
Some code portions could be simplified or removed if we were able to access the Aggregations dictionary info during the answer parsing.

For instance, in percentiles visitor, we add keyed and percentiles values to grab them back in the column name.
['A%percentile%25.0%50.0%99.0%False']=percentiles_array(fieldA, 25,50,99)']

The logic solution will be instead to only use the key and retrieve info needed in Aggregations info.

To do that, I would like to pass Aggregations dictionary in QueryData.

                var queryData = new QueryData(
                    elasticSearchDsl.KustoQL,
                    elasticSearchDsl.IndexName,
                    sortFields,
                    docValueFields,
                    elasticSearchDsl.HighlightText,
                    primaryAggregation);   <------ Add Aggregations dictionary in addition (or probably remove primaryAggregation as it's already contained in the dictionary).

But .. we come back to a previous conversation.
All model class are internal, it's not possible to add the aggregation dictionary except if I change the accessibility from internal to public.

So I suggest: 
- Update from internal to public all model class used in the request.
- Update from {get; set;} to {get; internal set:} all model class properties.

### DIFF
--- a/K2Bridge/ElasticQueryTranslator.cs
+++ b/K2Bridge/ElasticQueryTranslator.cs
@@ -114,7 +114,8 @@ namespace K2Bridge
                     sortFields,
                     docValueFields,
                     elasticSearchDsl.HighlightText,
-                    primaryAggregation);
+                    primaryAggregation,
+                    elasticSearchDsl.Aggregations);
 
                 if (elasticSearchDsl.Highlight != null)
                 {
--- a/K2Bridge/Models/QueryData.cs
+++ b/K2Bridge/Models/QueryData.cs
@@ -5,6 +5,7 @@
 namespace K2Bridge.Models
 {
     using System.Collections.Generic;
+    using K2Bridge.Models.Request.Aggregations;
 
     /// <summary>
     /// Represents the data needed to search for information in the database.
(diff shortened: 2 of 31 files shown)


tamirkamara:
Thank you for posting this @dupuyjs.

The way I'm looking at this is that this app has several layers, each with it's own responsibility:
- Request models & converters are suppose to represent the object model of the query in a form easy to process (by the next layer).
- Visitors are responsible of translating the ES query to ADX query by going through the request model tree.
- Response parser is responsible of taking the ADX response and converting it to ES style response by populating the response model tree.

I recognize that the request holders information beneficial for creating the response, however I think we should be careful not to breach (too much) the _separation of concerns_ principal. 
Having the response parser access a large object model like aggregations means that you ask the parser to understand a big part of the request model. I think that's not its responsibility... If the aggregation (or any other request layer data) is of interest to the parser, we should try to find a better way to make the necessary info available to it rather than pass along the entire object and let the parser find what it needs on its own.

dupuyjs:
@tamirkamara Thanks for the feedback. Sorry for not having replied before. I get your point, and I will make another suggestion to not change access modifiers. I suppose we can make an intermediate object with less details.

eladiw:
@dupuyjs , can we close this pr and copy relevant parts to the docs section?

dupuyjs:
Close this discussion. Will add this discussion in issues to investigate.

PR #94, opened 2021-12-17
feat: Date Range aggregation

tomconte:

* Implement Date Range bucket aggregation


K2Bridge/Utils/DateMathParser.cs
```diff
@@ -0,0 +1,82 @@
...
+
+    /// <content>
+    /// Parser for the Date Math expressions.
+    /// </content>
+    public class DateMathParser
+    {
+        public static string ParseDateMath(string expr)
+        {
+            // See Wiki or ES docs for details on date math syntax
+            var pattern = @"^([0-9-T:]+|now)\|*([0-9-+]*)([a-zA-Z]*)\/?([a-zA-Z]*)";
```
AsafMah, line 17: Sadly I think that date math is more complex than that.
According to https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#date-math
and https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-daterange-aggregation.html

It seems to support more date formats. The example they give have a "." as a seperator. I think this also has to do with the "Format" argument that the aggregation gets, but I'm not sure. Either way, this regex doesn't catch it.
My recommendation is to extract the date, and try parsing it with DateTime.Parse. And if you have a format, use that as ParseExact first.

K2Bridge/Visitors/DateRangeAggregationExpressionVisitor.cs
```diff
@@ -0,0 +1,40 @@
...
+            Ensure.IsNotNull(dateRangeExpression, nameof(dateRangeExpression));
+            EnsureClause.StringIsNotNullOrEmpty(dateRangeExpression.Field, nameof(dateRangeExpression.Field));
+
+            var fromExpr = dateRangeExpression.From != null ? DateMathParser.ParseDateMath(dateRangeExpression.From) : "''";
+            var toExpr = dateRangeExpression.To != null ? DateMathParser.ParseDateMath(dateRangeExpression.To) : "''";
+
+            var gteExpr = dateRangeExpression.From != null ? $"['{dateRangeExpression.Field}'] >= {fromExpr}" : null;
+            var ltExpr = dateRangeExpression.To != null ? $"['{dateRangeExpression.Field}'] < {toExpr}" : null;
+
+            dateRangeExpression.BucketNameKustoQL = $"strcat({fromExpr}, '_', {toExpr})";
```
AsafMah, line 29: Is this consistent with the naming of the normal range buckets?
Also, why strcat in kusto and not in c#?

AsafMah (changes requested):

K2Bridge/Utils/DateMathParser.cs
tomconte: +1 for the "." separator.

The "format" argument is meant for formatting results and is not used by Kibana. It does not relate to the date format in the range expression. ("it is also possible to specify a date format by which the from and to response fields will be returned.")

I am relying on Kusto to parse the date string, using `make_datetime()`, so what I need is to properly extract the whole date string and pass it on. Kusto understand "." as date separator for example, so that would work (e.g. `print make_datetime('2018.02.01 13:12:13');`).

K2Bridge/Visitors/DateRangeAggregationExpressionVisitor.cs
tomconte: The `strcat()` is on the Kusto side because the datetime is also generated there, using `make_datetime()`.
I am using underscore '_' to separate the bucket name from/to, since the dash '-' is used in the datetime format.

K2Bridge/Utils/DateMathParser.cs
tomconte: NB: I suppose I could "steal" the Regex from the .NET client...

https://github.com/elastic/elasticsearch-net/blob/2f93905e1e8e22c3bb03bd4865ecc315fe063675/src/Nest/CommonOptions/DateMath/DateMath.cs#L25

Or even use the .NET client to split the DateMath expression components?

AsafMah: If we already have a dependency on it it's probably a great idea

K2Bridge/Visitors/DateRangeAggregationExpressionVisitor.cs
AsafMah: Got it.

I think we should have some sort of unified interface.
Between this, numeric range, and percentiles, each one uses a different separator.

K2Bridge/Visitors/IVisitor.cs
```diff
@@ -133,6 +133,18 @@ internal interface IVisitor
...
+        /// <summary>
+        /// Accepts a given visitable object and builds a Kusto query.
+        /// </summary>
+        /// <param name="rangeAggregation">The Range aggregation to visit.</param>
+        void Visit(DateRangeAggregation dateRangeAggregation);
+
+        /// <summary>
+        /// Accepts a range expression, and builds the Kusto query.
+        /// </summary>
+        /// <param name="rangeAggregationExpression">The range aggregation expression.</param>
```
dupuyjs, line 145: dateRangeAggregationExpression instead rangeAggregationExpression

K2Bridge/Visitors/IVisitor.cs
```diff
@@ -133,6 +133,18 @@ internal interface IVisitor
...
         void Visit(QueryStringClause queryStringClause);
 
+        /// <summary>
+        /// Accepts a given visitable object and builds a Kusto query.
+        /// </summary>
+        /// <param name="rangeAggregation">The Range aggregation to visit.</param>
+        void Visit(DateRangeAggregation dateRangeAggregation);
+
+        /// <summary>
+        /// Accepts a range expression, and builds the Kusto query.
```
dupuyjs, line 143: /// Accepts a given visitable object and builds a Kusto query. <- To be aligned with other docstrings

K2Bridge/Visitors/DateRangeVisitor.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            // Part 2 is expansion and filtering of rows
+            dateRangeAggregation.KustoQL += $" | {KustoQLOperators.MvExpand} {EncodeKustoField(dateRangeAggregation.Key)} to typeof(string), ['_range_value']";
+            dateRangeAggregation.KustoQL += $" | {KustoQLOperators.Where} ['_range_value'] == true";
+
+            // Part 3 is the summarize part for metrics
+            dateRangeAggregation.KustoQL += $" | {KustoQLOperators.Summarize} {dateRangeAggregation.SubAggregationsKustoQL}{dateRangeAggregation.Metric} by {EncodeKustoField(dateRangeAggregation.Key)}";
+
+            // Order rows by key, and re-order columns by ascending order
+            // Make sure the aggregation column (with range names) is first
```
dupuyjs, line 72: Not needed any longer to have the primary aggregation column first. Now test within aggregates is done on the real column name == primary aggregation key.

K2Bridge/Utils/DateMathParser.cs
tomconte: We don't use it currently, we would need to add it: https://www.nuget.org/packages/NEST/

K2Bridge/Visitors/DateRangeVisitor.cs
```diff
@@ -0,0 +1,76 @@
...
+// See LICENSE file in the project root for full license information.
+
+namespace K2Bridge.Visitors
+{
+    using System.Globalization;
+    using K2Bridge.Models.Request.Aggregations;
+    using K2Bridge.Models.Response;
+
+    /// <content>
+    /// A visitor for the <see cref="RangeAggregation"/> element.
```
dupuyjs, line 12: DateRangeAggregation instead RangeAggregation

K2Bridge/Visitors/DateRangeAggregationExpressionVisitor.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace K2Bridge.Visitors
+{
+    using System;
+    using System.Text.RegularExpressions;
+    using K2Bridge.Models.Request;
+    using K2Bridge.Models.Request.Aggregations;
+
+    /// <content>
+    /// A visitor for the <see cref="RangeAggregationExpression"/> element.
```
dupuyjs, line 13: DateRangeAggregationExpression  instead RangeAggregationExpression

K2Bridge/Visitors/DateRangeAggregationExpressionVisitor.cs
tomconte: Right, I see that we now have `AggregationsConstants.MetadataSeparator` (currently set to '%'), I suppose it makes sense to use that.

K2Bridge/Models/Request/Aggregations/Bucket/DateRange/DateRangeAggregation.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// Gets or sets the keyed flag, indicating the bucket should be returned as a hash.
+        /// </summary>
+        [JsonProperty("keyed")]
+        public bool Keyed { get; set; }
+
+        /// <summary>
+        /// Gets or sets the time zone.
+        /// </summary>
+        [JsonProperty("time_zone")]
+        public string TimeZone { get; set; }
```
dupuyjs, line 39: Is TimeZone really used ? If not, we can perhaps add a workitem to implement TimeZone behavior for both DateRange and DateHistogram.

dupuyjs (changes requested):

K2Bridge/Utils/DateMathParser.cs
dupuyjs: Oh, didn't get this package was available. It could have helped before. As we are now independent, don't know if this is a good thing to introduce this dependency, it depends if using the new regex involves big changes or not. Let's you and @AsafMah decide.

K2Bridge/Visitors/DateRangeAggregationExpressionVisitor.cs
tomconte: Fixed to use MetadataSeparator.

K2Bridge/Utils/DateMathParser.cs
tomconte: I looked at the package, but it's a pretty big dependency to take and I am not even sure their API will let us do what we want (extract the distinct parts of the date math expression). I have pushed a change to use their regex.

K2Bridge/Models/Request/Aggregations/Bucket/DateRange/DateRangeAggregation.cs
tomconte: The time zone is ignored in general at the moment, it sounds like a good investigation user story to add to the backlog.

K2Bridge/Visitors/DateRangeAggregationExpressionVisitor.cs
tomconte: 👍🏻

K2Bridge/Visitors/IVisitor.cs
tomconte: 👍🏻

K2Bridge/Visitors/IVisitor.cs
tomconte: 👍🏻

K2Bridge/Models/Response/Aggregations/Bucket/DateRange/DateRangeBucket.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT license.
+// See LICENSE file in the project root for full license information.
+
+namespace K2Bridge.Models.Response.Aggregations
+{
+    using K2Bridge.JsonConverters;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// Terms bucket response.
```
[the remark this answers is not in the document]
tomconte: 👍🏻

K2Bridge/Models/Request/Aggregations/Bucket/DateRange/DateRangeAggregationExpression.cs
```diff
@@ -0,0 +1,50 @@
...
+// See LICENSE file in the project root for full license information.
+
+namespace K2Bridge.Models.Request.Aggregations
+{
+    using K2Bridge.JsonConverters;
+    using K2Bridge.Visitors;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// Range expression for aggregation.
```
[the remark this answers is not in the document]
tomconte: 👍🏻

K2Bridge/Visitors/DateRangeVisitor.cs
tomconte: OK, I removed the `project-reorder`.

K2Bridge/Visitors/DateRangeVisitor.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            Ensure.IsNotNull(dateRangeAggregation, nameof(RangeAggregation));
+            EnsureClause.StringIsNotNullOrEmpty(dateRangeAggregation.Metric, nameof(RangeAggregation.Metric));
+            EnsureClause.StringIsNotNullOrEmpty(dateRangeAggregation.Field, nameof(RangeAggregation.Field));
+
+            // Part 1:
+            // _data | extend ['_range'] = pack_array("range1", "range2", "range3"), ['_range_value']=pack_array(expr1, expr2, expr3)
+
+            // Start of query, until first pack_array()
+            dateRangeAggregation.KustoQL += $"_data | {KustoQLOperators.Extend} {EncodeKustoField(dateRangeAggregation.Key)} = {KustoQLOperators.PackArray}(";
```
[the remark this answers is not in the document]
tomconte: OK, too bad we won't be able to use `TrimEnd()` any more :-)

K2Bridge/Visitors/DateRangeVisitor.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            // Close the first pack_array() and start the second pack_array()
+            queryStringBuilder.Append($"), ['_range_value'] = {KustoQLOperators.PackArray}(");
+
+            // Insert range expressions
+            foreach (var range in dateRangeAggregation.Ranges)
+            {
+                if (string.IsNullOrEmpty(range.KustoQL))
+                {
+                    // This is then "open" range, -infinity to +infinity
```
dupuyjs, line 52: If needed, you can put elements of this section in an intermediate List<string>. Then use string.Join(',',myList) and put the final result in the stringBuilder. It avoids the additional coma.

K2Bridge/Visitors/DateRangeVisitor.cs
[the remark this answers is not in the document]
dupuyjs: Added a suggestion to avoid the additional coma.

K2Bridge/Visitors/DateRangeVisitor.cs
tomconte: This might be overkill, I think it's not that bad to call `Remove()` once or twice :-)

dupuyjs (approved):
Approved. Nice job.

K2Bridge/Utils/DateMathParser.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// <content>
+    /// Parser for the Date Math expressions.
+    /// </content>
+    public class DateMathParser
+    {
+        public static string ParseDateMath(string expr)
+        {
+            // See Wiki or ES docs for details on date math syntax
+            var pattern = @"^(?<anchor>now|.+?(?:\|\||$))(?<ranges>(?:(?:\+|\-)[^\/]*))?(?<rounding>\/(?:y|M|w|d|h|H|m|s))?$";
```
AsafMah, line 17: I don't really like this regex. It makes you do a lot of work to get the actual values from it.

If you want to keep it because we took it from somewhere else, that's ok (just do the comments in the rest of the file).
But I've took the liberty to rewrite this file to make it (in my opinion) nicer and more readable:

```c#
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT license.
// See LICENSE file in the project root for full license information.

namespace K2Bridge.Visitors
{
    using System.Text.RegularExpressions;

    /// <content>
    /// Parser for the Date Math expressions.
    /// </content>
    public class DateMathParser
    {
        public static string ParseDateMath(string expr)
        {
            // See Wiki or ES docs for details on date math syntax
            var rgx = new Regex(
                @"^
                        (?: (?<anchor>now) | (?: (?<anchor>.+?) (?:\|\||$)) )
                        (?: (?<rangeSign>[+-]) (?<rangeValue>\d+) (?<rangeUnit>[a-zA-Z]))?
                        (?: (?:\/ (?<rounding>y|M|w|d|h|H|m|s)))?", RegexOptions.IgnorePatternWhitespace);
            var match = rgx.Match(expr);

            var kexpr = string.Empty;

            if (!match.Success)
            {
                return kexpr;
            }

            // The date literal, ie.g. YYYY-MM-DD or "now"
            var anchor = match.Groups["anchor"].Value;
            if (anchor.Equals("now"))
            {
                kexpr = "now()";
            }
            else
            {
                kexpr = $"make_datetime('{anchor}')";
            }

            // Date operation, e.g. -1M, +3d, etc.
            if (match.Groups["rangeSign"].Success)
            {
                var sign = match.Groups["rangeSign"].Value == "-" ? "-" : string.Empty;
                var value = int.Parse(match.Groups["rangeValue"].Value);
                var unit = match.Groups["rangeUnit"].Value[0] switch {
                    'y' => "year",
                    'M' => "month",
                    'w' => "week",
                    'd' => "day",
                    'h' => "hour",
                    'H' => "hour",
                    'm' => "minute",
                    's' => "second",
                    _ => null,
                };

                if (unit != null)
                {
                    kexpr = $"datetime_add('{unit}', {sign}{value}, {kexpr})";
                }
            }

            // Rounding, e.g. /d, /h, /m, etc.
            if (match.Groups["rounding"].Success)
            {
                kexpr = match.Groups["rounding"].Value switch {
                    "y" => $"startofyear({kexpr})",
                    "M" => $"startofmonth({kexpr})",
                    "w" => $"startofweek({kexpr})",
                    "d" => $"startofday({kexpr})",
                    "h" => $"bin({kexpr}, 1h)",
                    "H" => $"bin({kexpr}, 1h)",
                    "m" => $"bin({kexpr}, 1m)",
                    "s" => $"bin({kexpr}, 1s)",
                    _ => kexpr,
                };
            }

            return kexpr;
        }
    }
}

```

So tell me what you think, I'm fine with either

K2Bridge/Utils/DateMathParser.cs
```diff
@@ -0,0 +1,84 @@
...
+
```
AsafMah, line 18: Technically there could be multiple range operations.
( see https://www.elastic.co/guide/en/elasticsearch/client/net-api/current/date-math-expressions.html  "Including multiple operations")

Do we want to support it?

K2Bridge/Visitors/DateRangeAggregationExpressionVisitor.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <inheritdoc/>
+        public void Visit(DateRangeAggregationExpression dateRangeExpression)
+        {
+            Ensure.IsNotNull(dateRangeExpression, nameof(dateRangeExpression));
+            EnsureClause.StringIsNotNullOrEmpty(dateRangeExpression.Field, nameof(dateRangeExpression.Field));
+
+            var fromExpr = dateRangeExpression.From != null ? DateMathParser.ParseDateMath(dateRangeExpression.From) : "''";
+            var toExpr = dateRangeExpression.To != null ? DateMathParser.ParseDateMath(dateRangeExpression.To) : "''";
+
+            var gteExpr = dateRangeExpression.From != null ? $"['{dateRangeExpression.Field}'] >= {fromExpr}" : null;
```
AsafMah, line 24: EncodeKustoField

AsafMah (changes requested):

K2Bridge.Tests.UnitTests/Visitors/DateMathParserTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        [TestCase("2018-01-01T14:42:30||-12h/m", ExpectedResult = "bin(datetime_add('hour', -12, make_datetime('2018-01-01T14:42:30')), 1m)")]
+        [TestCase("2018-01-01T14:42:30||+10h/s", ExpectedResult = "bin(datetime_add('hour', 10, make_datetime('2018-01-01T14:42:30')), 1s)")]
+        [TestCase("2018-01-01T14:42:30||-12h/s", ExpectedResult = "bin(datetime_add('hour', -12, make_datetime('2018-01-01T14:42:30')), 1s)")]
+        [TestCase("2018-01-01||+1d/h", ExpectedResult = "bin(datetime_add('day', 1, make_datetime('2018-01-01')), 1h)")]
+        [TestCase("2018-01-01||-2d/H", ExpectedResult = "bin(datetime_add('day', -2, make_datetime('2018-01-01')), 1h)")]
+        [TestCase("2018-01-01||+01d/m", ExpectedResult = "bin(datetime_add('day', 1, make_datetime('2018-01-01')), 1m)")]
+        [TestCase("2018-01-01||-012d/m", ExpectedResult = "bin(datetime_add('day', -12, make_datetime('2018-01-01')), 1m)")]
+        [TestCase("2018-01-01||+10d/s", ExpectedResult = "bin(datetime_add('day', 10, make_datetime('2018-01-01')), 1s)")]
+        [TestCase("2018-01-01||-1d/s", ExpectedResult = "bin(datetime_add('day', -1, make_datetime('2018-01-01')), 1s)")]
```
eladiw, line 64: 🚀

K2Bridge.Tests.UnitTests/Visitors/DateRangeVisitorTests.cs
```diff
@@ -0,0 +1,40 @@
...
+    using System.Collections.Generic;
+    using global::K2Bridge.Models.Request.Aggregations;
+    using global::K2Bridge.Tests.UnitTests.Visitors;
+    using global::K2Bridge.Visitors;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class DateRangeVisitorTests
+    {
+        [TestCase("2018-02-01", "2018-02-02", "2018-02-02", "2018-02-3", ExpectedResult = "_data | extend ['key'] = pack_array(strcat(make_datetime('2018-02-01'), '%', make_datetime('2018-02-02')),strcat(make_datetime('2018-02-02'), '%', make_datetime('2018-02-3'))), ['_range_value'] = pack_array(['dayOfWeek'] >= make_datetime('2018-02-01') and ['dayOfWeek'] < make_datetime('2018-02-02'),['dayOfWeek'] >= make_datetime('2018-02-02') and ['dayOfWeek'] < make_datetime('2018-02-3')) | mv-expand ['key'] to typeof(string), ['_range_value'] | where ['_range_value'] == true | summarize wibble by ['key'] | order by ['key'] asc")]
```
eladiw, line 16: dayOfWeek is a bit confusing since its a value between 1 to 7 (usually).
maybe change to time/date or something (but not to a keyword)

eladiw (approved):
looks good. just need to address @AsafMah  comments

K2Bridge/Utils/DateMathParser.cs
tomconte: Thanks Asaf, your version looks cleaner.

K2Bridge/Utils/DateMathParser.cs
```diff
@@ -0,0 +1,84 @@
...
+                    {
+                        'y' => "year",
+                        'M' => "month",
+                        'w' => "week",
+                        'd' => "day",
+                        'h' => "hour",
+                        'H' => "hour",
+                        'm' => "minute",
+                        's' => "second",
+                        _ => null,
```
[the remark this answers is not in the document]
tomconte: Done

K2Bridge/Utils/DateMathParser.cs
```diff
@@ -0,0 +1,84 @@
...
+                    {
+                        "y" => $"startofyear({kexpr})",
+                        "M" => $"startofmonth({kexpr})",
+                        "w" => $"startofweek({kexpr})",
+                        "d" => $"startofday({kexpr})",
+                        "h" => $"bin({kexpr}, 1h)",
+                        "H" => $"bin({kexpr}, 1h)",
+                        "m" => $"bin({kexpr}, 1m)",
+                        "s" => $"bin({kexpr}, 1s)",
+                        _ => kexpr,
```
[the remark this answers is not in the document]
tomconte: Done

K2Bridge/Utils/DateMathParser.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            Regex rgx = new Regex(pattern);
+            Match match = rgx.Match(expr);
+
+            var kexpr = string.Empty;
+
+            if (match.Success)
+            {
+                // The date literal, ie.g. YYYY-MM-DD or "now"
+                if (match.Groups["anchor"].Value.Equals("now"))
```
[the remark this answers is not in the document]
tomconte: Fixed by using your version of the code

K2Bridge/Utils/DateMathParser.cs
```diff
@@ -0,0 +1,84 @@
...
+                {
+                    kexpr = "now()";
+                }
+                else
+                {
+                    kexpr = $"make_datetime('{match.Groups["anchor"].Value.TrimEnd('|').TrimEnd('|')}')";
```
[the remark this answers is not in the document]
tomconte: Fixed by using your version of the code

K2Bridge/Utils/DateMathParser.cs
```diff
@@ -0,0 +1,84 @@
...
+                }
+
+                // Date operation, e.g. -1M, +3d, etc.
+                if (match.Groups["ranges"].Success)
+                {
+                    var rangeStr = match.Groups["ranges"].Value;
+                    var valueStr = rangeStr.Substring(0, rangeStr.Length - 1);
```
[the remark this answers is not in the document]
tomconte: Fixed by using your version of the code

K2Bridge/Visitors/DateRangeAggregationExpressionVisitor.cs
tomconte: Done

K2Bridge.Tests.UnitTests/Visitors/DateRangeVisitorTests.cs
tomconte: Done

K2Bridge/Visitors/DateRangeVisitor.cs
```diff
@@ -0,0 +1,80 @@
...
+            EnsureClause.StringIsNotNullOrEmpty(dateRangeAggregation.Metric, nameof(RangeAggregation.Metric));
+            EnsureClause.StringIsNotNullOrEmpty(dateRangeAggregation.Field, nameof(RangeAggregation.Field));
+
+            var queryStringBuilder = new StringBuilder();
+
+            // Part 1:
+            // _data | extend ['_range'] = pack_array("range1", "range2", "range3"), ['_range_value']=pack_array(expr1, expr2, expr3)
+
+            // Start of query, until first pack_array()
+            queryStringBuilder.Append($"_data | {KustoQLOperators.Extend} {EncodeKustoField(dateRangeAggregation.Key)} = {KustoQLOperators.PackArray}(");
```
eladiw, line 30: replace _data with its const (else where also, if applicable)

tomconte: 👍🏻

K2Bridge/Utils/DateMathParser.cs
```diff
@@ -0,0 +1,87 @@
...
+                {
+                    'y' => "year",
+                    'M' => "month",
+                    'w' => "week",
+                    'd' => "day",
+                    'h' => "hour",
+                    'H' => "hour",
+                    'm' => "minute",
+                    's' => "second",
+                    _ => throw new Exception("Invalid date math range unit."),
```
AsafMah, line 58: Throw a more specific exception. Maybe IllegalClause, but whatever you see fit

K2Bridge/Utils/DateMathParser.cs
tomconte: Right, so we could have `2018-02-01||+1d-1m` for example. I don't know how frequent that use case is, to be frank I missed it in the docs! 

I have pushed an attempt at parsing this using group captures, let me know what you think.

K2Bridge/Utils/DateMathParser.cs
tomconte: Done (in "Parse multiple ranges" commit)

K2Bridge/Utils/DateMathParser.cs
```diff
@@ -0,0 +1,90 @@
...
+            {
+                kexpr = "now()";
+            }
+            else
+            {
+                kexpr = $"make_datetime('{anchor}')";
+            }
+
+            // Date operation, e.g. -1M, +3d, etc.
+            if (match.Groups["rangeSign"].Success)
```
AsafMah, line 44: You don't need this if anymore, if there aren't any range expressions, the Captures.Count would be 0 and you won't get in the loop

K2Bridge/Utils/DateMathParser.cs
```diff
@@ -0,0 +1,90 @@
...
+                        'w' => "week",
+                        'd' => "day",
+                        'h' => "hour",
+                        'H' => "hour",
+                        'm' => "minute",
+                        's' => "second",
+                        _ => throw new IllegalClauseException("Invalid date math range unit."),
+                    };
+
+                    if (unit != null)
```
AsafMah, line 63: unit can't be null anymore

AsafMah (commented):
Small fixes

K2Bridge/Utils/DateMathParser.cs
tomconte: Right, fixed

K2Bridge/Utils/DateMathParser.cs
tomconte: Fixed

AsafMah (approved):
Looks good!

PR #95, opened 2021-12-19
Dynamic columns aggregations

AsafMah:
Adding support for aggregations for dynamic columns.

Aggregation functions in kusto (such as max, dcount, etc..) require a concrete type and can't work with just "dynamic",
so whenever we need it we query kusto for the field's actual type, and convert it with todouble/tostring etc.


K2Bridge/Visitors/QueryStringClauseVisitor.cs
```diff
@@ -110,11 +110,10 @@ private string CreateKqlFromLucenePhrase(QueryStringClause queryStringClause)
                 new QueryParser(
                     Lucene.Net.Util.Version.LUCENE_30,
                     queryStringClause.Default,
-                    analyzer)
-                {
-                    AllowLeadingWildcard = queryStringClause.Wildcard,
-                    LowercaseExpandedTerms = false,
-                };
+                    analyzer) {
```
eladiw, line 113: please revert change, so the pr will be minimal

K2Bridge/Models/Request/Aggregations/Aggregation.cs
```diff
@@ -14,7 +14,15 @@ namespace K2Bridge.Models.Request.Aggregations
...
-        public string Key { get; set; }
+        /// <summary>
+        ///  Key of the aggregation.
+        /// </summary>
+        public virtual string Key { get; set; }
+
+        /// <summary>
+        /// Aggregation field.
+        /// </summary>
+        public virtual string Field { get; set; }
```
dupuyjs, line 25: I understand the purpose of this change. And you're right, in the current implemented aggregations, all of them are sharing Field property. BUT, it's not the case for all bucket aggregations. For instance, if we implement Filters bucket aggregations, the field property is not supported (filters is used instead). It's also the case for additional ones not used by visualization (MultiTerms, Sampler, AdjancencyMatrix, ...). I suggest to rollback this change and keep Field as a top object property.

dupuyjs: For reference, only Metric Aggregations are all supporting Field property.

dupuyjs (changes requested):

K2Bridge/Models/Request/Aggregations/Bucket/Range/RangeAggregationExpression.cs
```diff
@@ -11,13 +11,8 @@ namespace K2Bridge.Models.Request.Aggregations
     /// <summary>
     /// Range expression for aggregation.
     /// </summary>
-    internal class RangeAggregationExpression : KustoQLBase, IVisitable
+    internal class RangeAggregationExpression : Aggregation, IVisitable
```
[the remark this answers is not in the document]
AsafMah: Ok, I rolled back

K2Bridge/Models/Request/Aggregations/Aggregation.cs
AsafMah: Let me know if this is ok

dupuyjs: Yes. All good for me. Thanks.

dupuyjs (approved):
lgfm. Great work.

K2Bridge/Visitors/RangeClauseVisitor.cs
```diff
@@ -53,7 +53,7 @@ public void Visit(RangeClause rangeClause)
             }
         }
 
-        private static void FillKqlQuery(RangeClause rangeClause, Func<string, string> valueConverter = null)
+        private void FillKqlQuery(RangeClause rangeClause, Func<string, string> valueConverter = null)
```
eladiw, line 56: why static was removed?

K2Bridge/Visitors/MetricAggregationVisitor.cs
```diff
@@ -72,8 +72,10 @@ public void Visit(PercentileAggregation percentileAggregation)
             var valuesForColumnNames = string.Join('%', percentileAggregation.Percents.ToList().Select(item => $"{item:0.0}"));
             var valuesForOperator = string.Join(',', percentileAggregation.Percents);
 
+            // We don't use EncodeKustoField on this key because it contains a '.' but isn't dynamic
```
eladiw, line 75: 👍

K2Bridge/Visitors/ElasticSearchDSLVisitor.cs
```diff
@@ -110,5 +112,28 @@ public void Visit(ElasticSearchDSL elasticSearchDSL)
...
+
+        private static bool IsFieldDynamic(string field) => field.Contains('.');
+
+        // Aggregations always need to be wrapped in a type
+        private string EncodeKustoField(MetricAggregation agg) => EncodeKustoField(agg.Field, true);
+
+        private string EncodeKustoField(string field, bool wrapDynamic = false)
+        {
+            var quoted = QuoteKustoField(field);
+            if (!IsFieldDynamic(field) || !wrapDynamic)
```
eladiw, line 126: when is wrapDynamic == true and IsFieldDynamic(field) = false?

K2Bridge/Startup.cs
```diff
@@ -80,12 +81,22 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddTransient<ITranslator, ElasticQueryTranslator>();
 
+            var dynamicSamplePercentage = GetConfigOptional<double>("dynamicSamplePercentage");
+            if (dynamicSamplePercentage is < 0 or > 100)
```
eladiw, line 85: what does dynamicSamplePercentage=0 means?
that we are not treating this field as dynamic? maybe it should be > 0

K2Bridge/Factories/BucketFactory.cs
```diff
@@ -181,8 +180,13 @@ public static void CreateAggregationColumns(Bucket bucket, DataRow row, ILogger
                     var columnName = column.ColumnName;
                     logger.LogTrace("Defining the value for {columnName}", columnName);
 
+                    if (!double.TryParse(row[columnName]?.ToString() ?? string.Empty, NumberStyles.Number, CultureInfo.InvariantCulture,  out var value))
```
eladiw, line 183: can we simplify this line ? ( ! ? ?? out)

K2Bridge.Tests.UnitTests/Visitors/RangeClauseVisitorTests.cs
```diff
@@ -168,7 +168,7 @@ private static string RangeClauseToKQL(RangeClause rangeClause)
 
         private static string DateRangeClauseToKQL(RangeClause rangeClause)
         {
-            var visitor = new ElasticSearchDSLVisitor(SchemaRetrieverMock.CreateMockDateSchemaRetriever());
+            var visitor = VisitorTestsUtils.CreateAndVisitRootVisitor("MyField", "date");
             VisitorTestsUtils.VisitRootDsl(visitor);
```
eladiw, line 172: l172 is redundant?

eladiw: (no need to visit twice)

K2Bridge.Tests.UnitTests/Visitors/RangeAggregationVisitorTests.cs
```diff
@@ -31,8 +30,30 @@ public string RangeVisit_WithAggregation_ReturnsValidResponse(double? from1, dou
...
+                Metric = "wibble",
+                Field = "dayOfWeek.A",
+                Key = "key",
+                Ranges = new List<RangeAggregationExpression>() {
+                    new RangeAggregationExpression { Field = "dayOfWeek.A", From = from1, To = to1 },
+                    new RangeAggregationExpression { Field = "dayOfWeek.A", From = from2, To = to2 },
+                },
+            };
+
+            var visitor = VisitorTestsUtils.CreateAndVisitRootVisitor("dayOfWeek.A", "double");
```
eladiw, line 56: l57 is redundant (twice visited)

eladiw (changes requested):
Looks good! some comments

K2Bridge/Visitors/RangeClauseVisitor.cs
AsafMah: Because `schemaRetriever` is a member in the class, so `EncodeKustoField` isn't static, so this can't be

K2Bridge/Startup.cs
AsafMah: You're right, it's now >0

K2Bridge/Factories/BucketFactory.cs
AsafMah: Is this better?

K2Bridge.Tests.UnitTests/Visitors/RangeAggregationVisitorTests.cs
AsafMah: Here it's different, the first line visits the root visitor and the second visits the aggregator

eladiw: ok, so why do we need to visit the root?

AsafMah: It initializes a bunch of stuff, including `schemaRetriever`. It's in most of the tests.

eladiw (approved):

PR #97, opened 2021-12-19
Update package versions

tamirkamara:

* Update all packages to their latest version
* Remove Serilog.Filters.Expressions as it's deprecated and doesn't seem to be used


AsafMah (approved):
lgtm

eladiw (approved):

PR #100, opened 2021-12-19
cherry pick pipe changes

eladiw:

* cherry pick pipe changes

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,7 +19,7 @@ stages:
 
   - job: security_analysis
     displayName: Security Analysis
-    condition: and(always(), not(variables['IS_DEMO_PIPELINE']))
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -19,7 +19,7 @@ stages:
 
   - job: security_analysis
     displayName: Security Analysis
-
+    condition: and(always(), not(variables['IS_DEMO_PIPELINE']))
```
tamirkamara, line 22: What does always() mean here? can't you just have the not(...) without the and?

tamirkamara (changes requested):

azure-pipelines.yml
eladiw: fixed

tamirkamara (approved):

AsafMah (approved):

PR #102, opened 2021-12-21
[Bug-15963] - Wrapping field names with ['']

eladiw:
In some cases, the field name was not wrapped with [' ']. 
In some edge cases this can lead to Kusto parsing failure, e.g. if the field name is a reserved word like 'datetime'.

* wrap the field name and fix all related tests

### DIFF
--- a/K2Bridge.Tests.End2End/QueryControllerErrorsTests.cs
+++ b/K2Bridge.Tests.End2End/QueryControllerErrorsTests.cs
@@ -19,7 +19,7 @@ namespace K2Bridge.Tests.End2End
                             ""root_cause"": [
                             {
                                 ""type"": ""SemanticException"",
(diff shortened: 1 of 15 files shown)


AsafMah:
Sadly this isn't good enough for dynamic field.
A field like `StormSummary.TotalDamages` needs to be quoted like `['StormSummary'].['TotalDamages']` and not `['StormSummary.TotalDamages']`

I can contribute this fix if you want

AsafMah (approved):

eladiw:
> Sadly this isn't good enough for dynamic field. A field like `StormSummary.TotalDamages` needs to be quoted like `['StormSummary'].['TotalDamages']` and not `['StormSummary.TotalDamages']`
> 
> I can contribute this fix if you want

Discussed f2f. will be taken care in your PR

dupuyjs (approved):
lgfm

PR #103, opened 2021-12-22
Update serialization model for aggregation response.

dupuyjs:

#### Main Changes
- Update serialization model for aggregation response. The previous model was not compliant to support standalone aggregations (for instance using max metrics without bucket aggregations). 
    - In addition, this new model will be able to support nested aggregation (sub bucket feature).
    - `ResponseElement` contains now an `AggregateDictionary` with `IAggregate` elements.
    - This dictionary contains different type of aggregates family:
        -  `ValueAggregate` for metric aggregation value.
        -  `PercentileAggregate` for percentile aggregation values.
        -  `BucketAggregate` for bucket aggregations.
    - `BucketAggregate` contains a collection of IBucket.
        - DateHistogramBucket for DateHistogram aggregation.
        - RangeBucket for Range aggregation.
        - TermsBucket for Terms aggregation.
    - All bucket object inherits from AggregateDictionary to contain a collection of aggregates previously described. 

- Update QueryData to include the primary key aggregation in addition to primary aggregation type.
- Review global parsing logic of aggregations to support metrics with no bucket aggregation.

#### Issues fixed

- Workitem 15347: Metrics with no bucket aggregation should return value.
- Workitem 15346: Count metric with no bucket aggregation should return total doc count.
- Workitem 15049: Expected json for ElasticResponseConstructor_OnValidInput_CreatesValidResponse is not valid.


K2Bridge.Tests.UnitTests/Visitors/ElasticSearchDSLVisitorTests.cs
```diff
@@ -17,7 +17,7 @@ namespace UnitTests.K2Bridge.Visitors
     public class ElasticSearchDSLVisitorTests
     {
         [TestCase(ExpectedResult =
-            "let _data = database(\"\").myindex | where (dayOfWeek == 1);\n(_data | limit 0 | as hits)")]
+            "let _data = database(\"\").myindex | where (dayOfWeek == 1);\n(_data | count | as hitsTotal);\n(_data | limit 0 | as hits)")]
```
eladiw, line 20:
```suggestion
            "let _data = database(\"\").myindex | where (dayOfWeek == 1);\n(_data | count | as hitsTotal);\n(_data | limit 0 | as hits)")]
```


why is hitTotal needed?

dupuyjs: It has been introduced by @tomconte when he worked on Ranges. I suppose it was related to overlapped ranges not computing the correct addition. Anyway, it also help now to retrieve the total count even there is no aggregation. (before, when you were performing just a count of the all dataset, it was returning only '1').

eladiw: why is it marked as being added now?

dupuyjs: If you check the code from ElasticSearchDSLVisitor, the **hitsTotal** was previsously added only in case of **aggregations.Count > 0**.  So, in this test, this block was not created.

It's not the case any longer, this hitsTotal block have been moved out and is called even there is no aggregation. This change has been done to catch the correct total count if we don't have any metric/bucket aggregation requested. It fixes the bug 15346. So **hitsTotal** is now added in the test.

dupuyjs: For reference: 

                // We will need the "true" hits count for some aggregations, e.g. Range
                // And this line must be added even there is no aggregation (default count metric)
                // KQL ==> (_data | count | as hitsTotal);
                queryStringBuilder.Append($"\n(_data | {KustoQLOperators.Count} | as hitsTotal);");

dupuyjs: @eladiw @AsafMah Don't hesitate to comment this PR this week. I can fix any issue you'll detect. Thanks.

K2Bridge/Models/Response/Aggregations/BucketAggregate.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    using System.Collections.Generic;
+    using K2Bridge.JsonConverters;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// Describes bucket aggregate response element.
+    /// </summary>
+    [JsonConverter(typeof(BucketAggregateConverter))]
+    public class BucketAggregate<TBucket> : IAggregate
```
AsafMah, line 15: Just a thought - do we need this to be generic?
It complicates the code without any real benefit (unless the serialization needs it?).
Like if instead of being generic, it had a `List<IKeyedbucket>`, and IKeyedBucket had an `object Key` property, you could use this class more easily,
for example in `BucketsAggregateConverter`, you would not need the generic method and could replace the if-else with one call.

But maybe there is a reason for this?

K2Bridge/Models/Response/ElasticResponse.cs
```diff
@@ -6,6 +6,7 @@ namespace K2Bridge.Models.Response
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
```
AsafMah, line 9: remvoe

K2Bridge.Tests.UnitTests/KustoDAL/KustoResponseParserTests.cs
```diff
@@ -137,15 +139,17 @@ public void ParseElasticResponse_WithHits_ReturnsElasticResponseWithHits()
             Assert.AreEqual(2, elasticResult.Hits.Hits.Count());
         }
 
-        [TestCase(nameof(DateHistogramAggregation))]
-        [TestCase(nameof(TermsAggregation))]
-        public void ParseElasticResponse_WithAggs_ReturnsElasticResponseWithAggs(string primaryAggregation)
+        [Test]
+        public void ParseElasticResponse_WithDateHistogramAggs_ReturnsElasticResponseWithAggs()
```
AsafMah, line 143: The terms agg test is gone?

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -0,0 +1,97 @@
...
+                        aggregateDictionary.Add(key, GetPercentileAggregate(column.ColumnName, columnMetadata, row, logger));
+                    }
+                }
+                else
+                {
+                    var key = column.ColumnName;
+                    logger.LogTrace($"Defining the value for {key}");
+
+                    var rowValue = (double)row[key];
+                    double? value = double.IsNaN(rowValue) ? null : rowValue;
```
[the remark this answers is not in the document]
dupuyjs: @AsafMah It's not the initial reason. When we target standalone aggregation (without any bucket), we get a similar KQL syntax: 

_data | summarize avg(field) -> There is no by grouping expression. 

Where _data has been previously filtered by time range. If the summarize is done on data out of range, we receive a valid 'aggs' table with one column and one row. And the cell contains a NaN value. Do you think we can avoid to receive NaN value by updating the KQL syntax ?

eladiw:
@dupuyjs  looks good. I will approve after the comments will be resolved

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <param name="row">The row to be added as aggregate.</param>
+        /// <param name="logger">ILogger object for logging.</param>
+        /// <param name="primaryKey">The primary aggregation key.</param>
+        public static void AddAggregates(this AggregateDictionary aggregateDictionary, DataRow row, ILogger logger, string primaryKey)
+        {
+            var columns = row.Table.Columns;
+
+            foreach (DataColumn column in columns)
+            {
+                if (column.ColumnName == BucketColumnNames.Count || column.ColumnName == primaryKey)
```
eladiw, line 33: name == count ?
are we expecting the name to be an int? if its intentional, lets add a comment

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    // Column Metadata (Separator %)
+                    // Structure: key%metric%value1%value2%keyed
+                    var metric = columnMetadata[1];
+
+                    if (metric == "percentile")
+                    {
+                        var key = columnMetadata[0];
+                        aggregateDictionary.Add(key, GetPercentileAggregate(column.ColumnName, columnMetadata, row, logger));
+                    }
```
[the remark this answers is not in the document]
eladiw: and log it

eladiw:
While testing the solution I tried using the min/max and average metrics with a datehistogram bucket. got an invalid kql:

let _data = database(\"xxxx\").kibana_sample_data_flights | where (['timestamp'] >= todatetime(\"2007-01-02T17:42:31.1690000Z\") and ['timestamp'] <= todatetime(\"2022-01-02T17:42:31.1690000Z\"));\n(_data | summarize count() by ['2'] = ['FlightDelay']\n| order by count_ desc\n| limit 5 | as aggs);\n(_data | count | as hitsTotal);\n(_data | limit 0 | as hits)

eladiw:
> While testing the solution I tried using the min/max and average metrics with a datehistogram bucket. got an invalid kql:
> 
> let _data = database("xxxx").kibana_sample_data_flights | where (['timestamp'] >= todatetime("2007-01-02T17:42:31.1690000Z") and ['timestamp'] <= todatetime("2022-01-02T17:42:31.1690000Z"));\n(_data | summarize count() by ['2'] = ['FlightDelay']\n| order by count_ desc\n| limit 5 | as aggs);\n(_data | count | as hitsTotal);\n(_data | limit 0 | as hits)

Also, need to check how this regression passed the unit tests and e2e tests

AsafMah:
@eladiw It seems to also be true for datetime fields.
We need to update the metric aggregations to support all types

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -0,0 +1,97 @@
...
+            var columns = row.Table.Columns;
+
+            foreach (DataColumn column in columns)
+            {
+                if (column.ColumnName == BucketColumnNames.Count || column.ColumnName == primaryKey)
+                {
+                    continue;
+                }
+
+                var columnMetadata = column.ColumnName.Split('%');
```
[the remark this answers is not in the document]
dupuyjs: Fixed. Let me know if it's ok for you.

K2Bridge/Models/Response/ElasticResponse.cs
dupuyjs: Fixed.

K2Bridge/Factories/AggregateFactory.cs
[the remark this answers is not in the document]
dupuyjs: @AsafMah: Actually I was wrong. The NaN test is not needed; the type returned is System.DBNull when we are out of range.  I am fixing the point.

dupuyjs:
> @eladiw It seems to also be true for datetime fields. We need to update the metric aggregations to support all types

Yes, I added support for datetime fields; other types are converted to double (this is the default behavior of elastic), int are returned as double. It's unfortunately complexified the serialization with the need to support value_as_string in case of timestamp usage.

K2Bridge/Factories/AggregateFactory.cs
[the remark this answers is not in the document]
dupuyjs: I throw now an InvalidOperationException. Are you ok with the exception type ? @eladiw I suppose the exception will be caught and logged by default, -or- do you I need to add an additional log step ?

K2Bridge/Factories/AggregateFactory.cs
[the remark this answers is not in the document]
dupuyjs: Fixed.

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        /// <summary>
+        /// Get <see cref="PercentileAggregate"> object from a given <see cref="DataRow"/>.
+        /// </summary>
+        /// <param name="columnName">The column name.</param>
+        /// <param name="columnMetadata">The column metadata.</param>
+        /// <param name="row">The row to be added as aggregate.</param>
+        /// <param name="logger">ILogger object for logging.</param>
+        /// <returns></returns>
+        private static PercentileAggregate GetPercentileAggregate(string columnName, string[] columnMetadata, DataRow row, ILogger logger)
```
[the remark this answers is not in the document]
dupuyjs: I added log into GetPercentileAggregate and GetValueAggregate. Do you think I need to log the returned value/values ?

K2Bridge/Factories/AggregateFactory.cs
dupuyjs: It means we check if ColumName == "count_" or ColumnName == (a string primary key). There is no more mix between index and real column names. All the values are string.

dupuyjs: For reference, we were supposing before the primary aggregation column was the first one. Now I am getting the real name of the column (the primary key string).

K2Bridge/JsonConverters/BucketAggregateConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            if (value is BucketAggregate<DateHistogramBucket> dateHistogramAggregate)
+            {
+                serializer.Serialize(writer, dateHistogramAggregate.Buckets);
+            }
+
+            if (value is BucketAggregate<TermsBucket> termsAggregate)
+            {
+                serializer.Serialize(writer, termsAggregate.Buckets);
+            }
```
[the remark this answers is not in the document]
dupuyjs: Fixed.

K2Bridge/JsonConverters/DateHistogramBucketConverter.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// This converter serializes <see cref="DateHistogramBucket"/> to Elasticsearh response json format.
+    /// </summary>
+    internal class DateHistogramBucketConverter : WriteOnlyJsonConverter
+    {
+        /// <inheritdoc/>
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            writer.WriteStartObject();
+
+            if (value is DateHistogramBucket dateHistogramBucket)
```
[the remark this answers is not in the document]
dupuyjs: Fixed.

K2Bridge/JsonConverters/TermsBucketConverter.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// This converter serializes <see cref="TermsBucket"/> to Elasticsearh response json format.
+    /// </summary>
+    internal class TermsBucketConverter : WriteOnlyJsonConverter
+    {
+        /// <inheritdoc/>
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            writer.WriteStartObject();
+
+            if (value is TermsBucket termsBucket)
```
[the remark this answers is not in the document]
dupuyjs: Fixed.

K2Bridge/Models/Response/Aggregations/BucketAggregate.cs
dupuyjs: Yes. I agree. Initially, I wanted to avoid using 'object' to ensure we check the type. And, the second point, it seems some elastic search buckets are not always inheriting KeyedBucket. Anyway, as I am not sure we will implement them, we can remove this complexity at the moment. Fixed.

K2Bridge.Tests.UnitTests/KustoDAL/KustoResponseParserTests.cs
dupuyjs: Yes. I created a new dedicated test for terms aggregation.

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -0,0 +1,228 @@
...
+            foreach (DataColumn column in columns)
+            {
+                if (column.ColumnName == BucketColumnNames.Count || column.ColumnName == primaryKey)
+                {
+                    continue;
+                }
+
+                // Column Metadata (Separator %)
+                // Structure: key%metric%value1%value2%keyed
+                const char Seperator = '%';
```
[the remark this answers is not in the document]
dupuyjs: Yes. True.

eladiw (approved):
lgtm

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -0,0 +1,228 @@
...
+
+namespace K2Bridge.Factories
+{
+    using System;
+    using System.Linq;
+    using System.Data;
+    using System.Globalization;
+    using K2Bridge.Models.Response;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
```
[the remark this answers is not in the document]
dupuyjs: Fixed.

K2Bridge/Factories/AggregateFactory.cs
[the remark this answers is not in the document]
dupuyjs: Fixed.

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -0,0 +1,230 @@
...
+                    {
+                        Percentile = double.Parse(percent, CultureInfo.InvariantCulture),
+                        Value = null,
+                    };
+
+                    percentileAggregate.Values.Add(percentileItem);
+                }
+            }
+
+            logger.LogTrace("Percentile aggregate returned for {0}: {1}", columnName, percentileAggregate);
```
AsafMah, line 197:
```suggestion
            logger.LogTrace("Percentile aggregate returned for {}: {}", columnName, percentileAggregate);
```
(And in the other one too)

AsafMah (approved):
Overall looks good

K2Bridge/Factories/AggregateFactory.cs
dupuyjs: Fixed.

PR #104, opened 2021-12-23
fix: Normalize aggregate metric value

dupuyjs:

- Remove TODO comment for workitem 15247 and update the normalization method name.

For reference, all metric aggregations (ValueAggregate) return 'value' except Percentile (PercentileAggregate) that will return 'values'. Additional support for 'values' is not needed in this method name as Percentile has its own dedicated tests. And, this normalization feature cannot be reused for percentiles as difference between values is too important. An issue is still opened for this last point (Bug 15795).


eladiw (approved):

AsafMah (approved):
Makes sense

PR #107, opened 2021-12-29
Add MatchAll query

AsafMah:
Added MatchAll query.
This solves the issue when an error pops up when you just type "*"

### DIFF
--- a/K2Bridge.Tests.UnitTests/Visitors/LuceneNet/LuceneMatchAllDocsVisitorTests.cs
+++ b/K2Bridge.Tests.UnitTests/Visitors/LuceneNet/LuceneMatchAllDocsVisitorTests.cs
@@ -1,0 +1,58 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// See LICENSE file in the project root for full license information.
+
+namespace UnitTests.K2Bridge.Visitors.LuceneNet
+{
+    using System;
+    using global::K2Bridge.Models.Request.Queries;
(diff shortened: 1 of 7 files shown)


K2Bridge.Tests.UnitTests/Visitors/LuceneNet/LuceneMatchAllDocsVisitorTests.cs
```diff
@@ -0,0 +1,58 @@
...
+            var query = new LuceneMatchAllDocsQuery {
+                LuceneQuery = null,
+            };
+            var visitor = new LuceneVisitor();
+            Assert.That(
+                () => visitor.Visit(query),
+                Throws.TypeOf<IllegalClauseException>());
+        }
+
+        [TestCase(ExpectedResult = null)]
```
eladiw, line 38: why expected result is null?

eladiw:
looks good. one question

K2Bridge.Tests.UnitTests/Visitors/LuceneNet/LuceneMatchAllDocsVisitorTests.cs
AsafMah: See K2Bridge/Visitors/QueryStringClauseVisitor.cs.

Match all matches everything, so we don't need to add anything to the kusto query.
I can replace it with String.Empty if you prefer

eladiw (approved):
lgtm

YannickBrombach:
lgtm

YannickBrombach (approved):
lgtm

PR #108, opened 2022-01-03
Fix - Return empty results set when invalid field is used

eladiw:

* Instead of 500 Internal server error, return empty results set when invalid field is used

### DIFF
--- a/K2Bridge.Tests.End2End/ParallelApiTest.cs
+++ b/K2Bridge.Tests.End2End/ParallelApiTest.cs
@@ -33,6 +33,13 @@ namespace K2Bridge.Tests.End2End
 
         [Test]
         [Description("/_msearch Kibana aggregation query returning zero results")]
+        public void CompareElasticKusto_WhenMSearchContainsInvalidFieldNameZeroResults_ResponsesAreEquivalent()
+        {
+            ParallelQuery($"{FLIGHTSDIR}/MSearch_Text_Contains_Invalid_Field.json", minResults: 0);
+        }
+
+        [Test]
+        [Description("/_msearch Kibana aggregation query returning zero results")]
         public void CompareElasticKusto_WhenMSearchZeroResults_ResponsesAreEquivalent()
         {
             ParallelQuery($"{FLIGHTSDIR}/MSearch_ZeroResults_Equivalent.json", minResults: 0);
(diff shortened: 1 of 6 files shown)


tamirkamara (commented):
So the Kibana server normally returns the same response in an error or in an empty set?
How does the user know the difference?

eladiw:
> So the Kibana server normally returns the same response in an error or in an empty set? How does the user know the difference?


For invalid field name, Kibana returns an empty result, the user can't tell.
for other kusto errors I left the exception, so if something goes wrong (e.g. an invalid generated kql), we still get an error

K2Bridge.Tests.End2End/ParallelApiTest.cs
```diff
@@ -31,6 +31,13 @@ public void CompareElasticKusto_WhenClusterInfo_ResponsesAreEquivalent()
             AssertJsonIdentical(k2.Result, es.Result);
         }
 
+        [Test]
+        [Description("/_msearch Kibana aggregation query returning zero results")]
+        public void CompareElasticKusto_WhenMSearchContainsInvalidFieldNameZeroResults_ResponsesAreEquivalent()
+        {
+            ParallelQuery($"{FLIGHTSDIR}/MSearch_Text_Contains_Invalid_Field.json", minResults: 0);
```
YannickBrombach, line 38: maxResults?

K2Bridge/KustoDAL/KustoQueryExecutor.cs
```diff
@@ -110,10 +110,31 @@ public async Task<(TimeSpan timeTaken, IDataReader reader)> ExecuteQueryAsync(Qu
...
                 return (timeTaken, dataReader);
             }
+            catch (Kusto.Data.Exceptions.SemanticException ex)
+            {
+                Logger.LogError(ex, "Semantic exception - Failed to execute query.");
+
+                // If it was a semantic error AND it is about an invalid field, ignore
+                // the error and return an empty results set, as this is the behaviour of
+                // Kibana
+                if (!ex.Message.Contains(
```
AsafMah, line 120: I don't like that we have to check the exception text, because that seems like something that might change one day,
but I don't have a better way.

With that being said, I think that rather checking on the entire message (which also contains the query), you should check on `ex.SemanticErrors` or `ex.ErrorMessage`

PR #109, opened 2022-01-05
Reuse query table names across visitors and response parser

tamirkamara:

* Create a common-single place for the table names used in queries which will remove the need to hard-code them while creating the query expressions (in visitors) and while parsing the resultset.
* A bunch of tiny fixes to remove lint warnings.


tomconte (approved):
Looks good!

eladiw (approved):

PR #113, opened 2022-01-07
Feature - Managed identity

Mimetis:
# USING K2BRIDGE WITH MANAGED IDENTITY WITHIN AKS

We can remove all secrets stored in any app settings within **K2Bridge**, if we are relying on the **Managed Identity** mechanism provided in **Azure**.

## Requirements

Since using **Managed Identity** is only available once the application is actually deployed, we can rely locally on:
- **Local identity**, using the **Azure CLI Token**
- Continuing to use an **Service Principal** issued from the Application Azure AD Token

Since we already have the second solution in place, we can use both solutions:
- Using **Managed Identity** when solution is deployed in Azure.
- Using the `ClientId` / `ClientSecret` locally when working / debugging on our local machine.

The Pull Request will bring the managed identity choice when building the ADX connection string.

The rule is straightforward:
- When the settings `aadClientId` (or `aadClientSecret`) is not available, the **K2Bridge** application will use the **Managed Identity** authentication.
- When both `aadClientId` and `aadClientSecret` are available, **K2Bridge** will use the **Service Principal** authentication.

``` csharp
KustoConnectionStringBuilder conn;

if (string.IsNullOrEmpty(connectionDetails.AadClientId) || string.IsNullOrEmpty(connectionDetails.AadClientSecret))
    conn = new KustoConnectionStringBuilder(connectionDetails.ClusterUrl, connectionDetails.DefaultDatabaseName)
        .WithAadSystemManagedIdentity();
else
    conn = new KustoConnectionStringBuilder(connectionDetails.ClusterUrl, connectionDetails.DefaultDatabaseName)
        .WithAadApplicationKeyAuthentication(connectionDetails.AadClientId, connectionDetails.AadClientSecret,
            connectionDetails.AadTenantId);
```

## Azure Kubernetes Service

**K2Bridge** is using an aks cluster to deploy everything.
We will need a managed identity bind to the application (and POD) to be able to use the managed identity authentication

**Note** : AKS with Managed Identity : [Use managed identities in Azure Kubernetes Service](https://docs.microsoft.com/en-us/azure/aks/use-managed-identity)

### Push to docker

First of all, we need an application deployed in the **Azure Container Registry**:

``` bash
docker tag kustomanagedidentity:latest spertus.azurecr.io/kustomanagedidentity:latest
docker push spertus.azurecr.io/kustomanagedidentity:latest
```

### AKS

Create **AKS** Cluster with managed identity:

``` bash
az aks create -g rgkusto --name bridgeaks --enable-managed-identity --node-count 2 --generate-ssh-keys --attach-acr spertus

```
Once deployed, get the **CREDENTIALS**, and get the **AKS AGENT POOL IDENTITY**, and attach **ACR**:

``` bash
az aks get-credentials -g rgkusto -n bridgeaks
az aks show -g rgkusto -n bridgeaks --query "identity"

# We can't attach ACR during AKS creation, since we are using managed identity.
# Then we attach ACR after initial AKS creation
az aks update -g rgkusto -n bridgeaks --attach-acr spertus
```

### DEPLOYING THE K2BRIDGE APPLICATION

Get the **AGENTPOOL IDENTITY**:

``` bash
az aks show -g rgkusto -n bridgeaks --query "identityProfile"

{
  "kubeletidentity": {
    "clientId": "3ddc45c4-wwww-xxxxx-yyyyy-xxxxxxxxxxxxx",
    "objectId": "....",
    "resourceId": "/subscriptions/6d854ccd-e85d-4a3e-baee-af4f95a56211/resourcegroups/MC_rgkusto_bridgeaks_northeurope/providers/Microsoft.ManagedIdentity/userAssignedIdentities/bridgeaks-agentpool"
  }
}
```

Then add the **ClientId** to the permission in **ADX** (section **Permissions** from **Azure Portal**)

Then deploy the application in AKS

``` yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: kustomanagedidentity
  labels:
    app: kustomanagedidentity
spec:
  replicas: 1
  selector:
    matchLabels:
      app: kustomanagedidentity
  template:
    metadata:
      labels:
        app: kustomanagedidentity
    spec:
      containers:
        - name: kustomanagedidentity
          image: spertus.azurecr.io/kustomanagedidentity:latest
          imagePullPolicy: IfNotPresent
          ports:
            - containerPort: 80
              protocol: TCP
          env:
            - name: ASPNETCORE_URLS
              value: http://+:80
---
apiVersion: v1
kind: Service
metadata:
  name: kustomanagedidentity-svc
  labels:
    app: kustomanagedidentity-svc
spec:
  type: LoadBalancer
  ports:
    - port: 80
      targetPort: 80
      protocol: TCP
  selector:
    app: kustomanagedidentity
```

``` bash
kublect apply -f aks
```


AsafMah:
Hey, you accidentally made a PR for "master" instead of "dev-vnext", it makes it impossible to review, please fix

eladiw (changes requested):
Few comments:
a) the documentation is great, lets add it to the docs section?
b) what about Terraform  templates update to deploy with managed identity?

K2Bridge/KustoDAL/KustoQueryExecutor.cs
```diff
@@ -57,13 +57,26 @@ internal class KustoQueryExecutor : IQueryExecutor
...
-            var conn = new KustoConnectionStringBuilder(
-                connectionDetails.ClusterUrl,
-                connectionDetails.DefaultDatabaseName)
-                .WithAadApplicationKeyAuthentication(
-                    connectionDetails.AadClientId,
-                    connectionDetails.AadClientSecret,
-                    connectionDetails.AadTenantId);
+            KustoConnectionStringBuilder conn;
+
+            if (string.IsNullOrEmpty(connectionDetails.AadClientId) ||
```
AsafMah, line 62: I'm not a big fan of this approach. If someone forgets to have a secret, rather than informing them we will silently do Managed Identity auth.

I think we should add an explicit option in the config for the type of the authentication, and then throw if keys are missing.

Mimetis: Make sense.
I will introduce the `useManagedIdentity` bool variable in the **appsettings.json**.
The default value will be false to be backward compatible.

Mimetis:
> Few comments: a) the documentation is great, lets add it to the docs section? b) what about Terraform templates update to deploy with managed identity?

The terraform version used in the terraform templates is `1.38.0`
This version is not allowing us to use the property `identity`

``` yaml
  identity {
    type = "SystemAssigned"
  }
```
Upgrading terraform to the last version `2.91.0` breaks a lot of things in the terraform templates.

**I would rather prefer to not touch it for now, since I don't have enough knowledge on the project to handle all the side effects.**

That being said, I've create a small sample to be able to work on a new flag `aks_use_managed_identity` to allow the user to choose between having a client / secret or a managed identity for the aks cluster

### aks_use_managed_identity

About the flag `aks_use_managed_identity` in terraform.
It can be tricky to make a boolean condition on a whole section inside the aks resource.
So far the only way to make it running is to use the [dynamic](https://www.terraform.io/language/expressions/dynamic-blocks) statement:

So far, we can have a variable in the **variables.tf** file:

``` yaml
variable "aks_use_managed_identity" {
  type = bool
}
```

And then using the dynamic statement to have on or the other section:

``` yaml
resource "azurerm_kubernetes_cluster" "aks" {
  name                = var.aks_name

  dynamic "service_principal" {
    iterator = sp
    for_each = var.aks_use_managed_identity ? [] : [{}] false
    content {
      client_id     = var.aks_sp_client_id
      client_secret = var.aks_sp_client_secret
    }
  }

  dynamic "identity" {
    iterator = sp
    for_each = var.aks_use_managed_identity ? [{}] : [] false
    content {
      type = "SystemAssigned"
    }
  }
```
That being said, we still have to manage the default value for client / secret, that are still mandatory, and upgrading the terraform provider, and updating the devops pipeline

AsafMah (approved):
Looks good from my end,
but sort everything out with the terraform with @eladiw  before merging

eladiw (changes requested):
looks good. 
I am missing:
a) documentation
b) a new issue to update the existing terraform template to support this (mention it requires tf version upgrade)

Mimetis:
@eladiw Documentation has been migrated to the devops wiki

Mimetis:
I've updated the documentation partially, since it could not be completed until we have:
- Update the **helm chart** to support the `useManagedIdentity` settings
- Update the **helm chart** to ensure that `aadClientSecret` is now optional (if `useManagedIdentity` is set to true)

Not part of the doc, but still needed:
- Update the **terraform** module to the last version (to support aks creation with managed identity)
- Update the **terraform** module to allow user to use either **managed identity** or **service principal**. (see previous comment on the new `aks_use_managed_identity` variable and the dynamic implementation
- Update the **terraform** module to ensure that `aks_sp_client_id`, `aks_sp_object_id`, `aks_sp_client_secret` are optional if `aks_use_managed_identity` is set to true

docs/installation.md
```diff
@@ -2,18 +2,40 @@
 
 ## Requirements
 
-* AKS cluster ([create an AKS cluster instructions](https://docs.microsoft.com/en-us/azure/aks/kubernetes-walkthrough-portal#create-an-aks-cluster)) or any other Kubernetes cluster (version 1.14 or newer - tested and verified up to version 1.16. A minimum of 3 node count).
+### AKS
+
+You need an AKS cluster ([create an AKS cluster instructions](https://docs.microsoft.com/en-us/azure/aks/kubernetes-walkthrough-portal#create-an-aks-cluster)) or any other Kubernetes cluster (version 1.14 or newer - tested and verified up to version 1.16. and a minimum of 3 node count).
```
eladiw, line 7:
```suggestion
You need an AKS cluster ([create an AKS cluster instructions](https://docs.microsoft.com/en-us/azure/aks/kubernetes-walkthrough-portal#create-an-aks-cluster)) or any other Kubernetes cluster (version 1.21.2 or newer - tested and verified with version 1.21.2. and a minimum of 3 node count).  
```

eladiw (approved):
lgtm.

PR #114, opened 2022-01-07
Feature: Filters aggregation

tomconte:

* Implement Filters aggregation

Final generated query and example ADX output:
image

### DIFF
--- a/K2Bridge/JsonConverters/FiltersBucketConverter.cs
+++ b/K2Bridge/JsonConverters/FiltersBucketConverter.cs
@@ -30,10 +30,10 @@ namespace K2Bridge.JsonConverters
                 serializer.Serialize(writer, filtersBucket.Key);
             }
 
-            foreach (KeyValuePair<string, IAggregate> aggregate in filtersBucket)
+            foreach (var (key, aggregate) in filtersBucket)
             {
-                writer.WritePropertyName(aggregate.Key);
-                serializer.Serialize(writer, aggregate.Value);
+                writer.WritePropertyName(key);
+                serializer.Serialize(writer, aggregate);
             }
 
             writer.WriteEndObject();
--- a/K2Bridge/JsonConverters/FiltersBucketConverter.cs
+++ b/K2Bridge/JsonConverters/FiltersBucketConverter.cs
@@ -1,0 +1,42 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// See LICENSE file in the project root for full license information.
+
+namespace K2Bridge.JsonConverters
+{
+    using System.Collections.Generic;
+    using K2Bridge.JsonConverters.Base;
+    using K2Bridge.Models.Response.Aggregations;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// This converter serializes <see cref="FiltersBucket"/> to Elasticsearh response json format.
+    /// </summary>
+    internal class FiltersBucketConverter : WriteOnlyJsonConverter
+    {
+        /// <inheritdoc/>
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            var filtersBucket = (FiltersBucket)value;
+
+            writer.WriteStartObject();
+
+            writer.WritePropertyName("doc_count");
+            serializer.Serialize(writer, filtersBucket.DocCount);
+
+            if (filtersBucket.Key is not null)
+            {
+                writer.WritePropertyName("key");
+                serializer.Serialize(writer, filtersBucket.Key);
+            }
+
+            foreach (KeyValuePair<string, IAggregate> aggregate in filtersBucket)
+            {
+                writer.WritePropertyName(aggregate.Key);
+                serializer.Serialize(writer, aggregate.Value);
+            }
+
+            writer.WriteEndObject();
+        }
+    }
+}
--- a/K2Bridge/Visitors/Aggregations/FiltersAggregationVisitor.cs
+++ b/K2Bridge/Visitors/Aggregations/FiltersAggregationVisitor.cs
@@ -19,6 +19,8 @@ namespace K2Bridge.Visitors
         {
             Ensure.IsNotNull(filtersAggregation, nameof(FiltersAggregation));
 
+            var expandColumn = EncodeKustoField("_filter_value");
+
             var queryStringBuilder = new StringBuilder();
 
             // Part 1:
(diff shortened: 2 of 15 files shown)


eladiw (approved):
Looks good!

AsafMah:
Before diving in the code, I tried it manually, and it seems that the summerize is always empty:
Text=let _data = database("demo_db").flights | where (['timestamp'] >= todatetime("2007-01-09T10:15:48.8550000Z") and ['timestamp'] <= to
datetime("2022-01-09T10:15:48.8550000Z"));
(_data | summarize  | as aggs);
(_data | count | as hitsTotal);
(_data | limit 0 | as hits)
SemanticErrors=Invalid summarize operator: missing both aggregates and group-by expressions

I tried with demo_db:flights
With queries like `DestWeather:Rain` and `Cancelled:true`

K2Bridge/JsonConverters/FiltersBucketConverter.cs
```diff
@@ -0,0 +1,42 @@
...
+            writer.WritePropertyName("doc_count");
+            serializer.Serialize(writer, filtersBucket.DocCount);
+
+            if (filtersBucket.Key is not null)
+            {
+                writer.WritePropertyName("key");
+                serializer.Serialize(writer, filtersBucket.Key);
+            }
+
+            foreach (KeyValuePair<string, IAggregate> aggregate in filtersBucket)
```
AsafMah, line 33: Could simplify to 
```c#
            foreach (var (key, aggregate) in filtersBucket)
            {
                writer.WritePropertyName(key);
                serializer.Serialize(writer, aggregate);
            }
```

K2Bridge/Visitors/Aggregations/FiltersAggregationVisitor.cs
```diff
@@ -0,0 +1,68 @@
...
+            }
+
+            // Remove final comma
+            queryStringBuilder.Remove(queryStringBuilder.Length - 1, 1);
+
+            // Close the first pack_array() and start the second pack_array()
+            queryStringBuilder.Append($"), ['_filter_value'] = {KustoQLOperators.PackArray}(");
+
+            // Insert filters expressions
+            foreach (var filter in filtersAggregation.Filters)
```
AsafMah, line 43:
````c#
            foreach (var (_, value) in filtersAggregation.Filters)
            {
                value.BoolQuery.Accept(this);
                queryStringBuilder.Append($"{value.BoolQuery.KustoQL},");
            }
````

File: K2Bridge/Visitors/Aggregations/FiltersAggregationVisitor.cs [hunk 3]
```diff
@@ -0,0 +1,68 @@
...
+            {
+                filter.Value.BoolQuery.Accept(this);
+                queryStringBuilder.Append($"{filter.Value.BoolQuery.KustoQL},");
+            }
+
+            // Remove final comma
+            queryStringBuilder.Remove(queryStringBuilder.Length - 1, 1);
+
+            // End part 1
+            queryStringBuilder.Append(")");
```
AsafMah, line 53: ')'

AsafMah (approved):
Overall looks great! some minor styling points

One thing, is that the visitor here shares a lot of code with ther visitor in the date range pr, after they're merged, can we reduce the repetition?

AsafMah (commented):
removing approve for now

K2Bridge/Visitors/Aggregations/FiltersAggregationVisitor.cs
```diff
@@ -0,0 +1,68 @@
...
+            foreach (var filter in filtersAggregation.Filters)
+            {
+                queryStringBuilder.Append($"'{filter.Key}',");
+            }
+
+            // Remove final comma
+            queryStringBuilder.Remove(queryStringBuilder.Length - 1, 1);
+
+            // Close the first pack_array() and start the second pack_array()
+            queryStringBuilder.Append($"), ['_filter_value'] = {KustoQLOperators.PackArray}(");
```
tamirkamara, line 40: I think you can/should extract **['_filter_value']** into a var (with EncodeKustoField) as you use it 3 times here...

eladiw:
Tried it. worked well for 1 and 2 filters. when added the third it crashed:

image

eladiw (changes requested):
looks good. Please verify the case of 3 filters...

AsafMah:
@eladiw It's not 3 filters, it's the * filter that doesn't work.

AsafMah:
That's because the MatchAll I added doesn't have any kql with it, and array_pack (and I assume other functions) expect a condition.

So to not make it a special case, we need to change in
QueryStringClauseVisitor.cs
from:
```
                case QueryStringClause.Subtype.MatchAll:
                    // Match all returns everything, so we don't need a query
                    break;
```
to:
```
                case QueryStringClause.Subtype.MatchAll:
                    queryStringClause.KustoQL = "true";
                    break;
```

Do you want to do it here, or I'll make another PR?

eladiw:
> @eladiw It's not 3 filters, it's the * filter that doesn't work.

Not only, any other string also fails

eladiw:
> That's because the MatchAll I added doesn't have any kql with it, and array_pack (and I assume other functions) expect a condition.
> 
> So to not make it a special case, we need to change in QueryStringClauseVisitor.cs from:
> 
> ```
>                 case QueryStringClause.Subtype.MatchAll:
>                     // Match all returns everything, so we don't need a query
>                     break;
> ```
> 
> to:
> 
> ```
>                 case QueryStringClause.Subtype.MatchAll:
>                     queryStringClause.KustoQL = "true";
>                     break;
> ```
> 
> Do you want to do it here, or I'll make another PR?

Separate PR, IMO

AsafMah:
@tomconte @eladiw 
The error was because a filter was set to kql instead of lucene.

Now we get an error if the filter doesn't match anything, can you check it out?

eladiw:
Verified. 
2+ filters are working correctly. I had 'kql' syntax turned on, which is not supported.
however, I noticed that if you filter too much and get zero results, then an 'undefined bucket' error is thrown.
image

K2Bridge/JsonConverters/FiltersBucketConverter.cs
tomconte: 👍🏻

K2Bridge/Visitors/Aggregations/FiltersAggregationVisitor.cs
tomconte: 👍🏻

K2Bridge/Visitors/Aggregations/FiltersAggregationVisitor.cs
tomconte: Done

tomconte:
Regarding the error when a filter does not match; in this case, the Kusto query does not return any row for the bucket, so when the response is assembled, the bucket is completely missing and Kibana does not like that.

Screenshot 2022-01-10 at 16 15 55

I think I might need to tweak the Kusto query so it returns a summarize row even if the expression in the array does not match any rows...

K2Bridge/Visitors/Aggregations/FiltersAggregationVisitor.cs
tomconte: 👍🏻

PR #115, opened 2022-01-09
Add editorconfig file

tamirkamara:

* Use editorconfig to align on file standard (encoding, file ends, etc.), as well as some of the styling roles.
Note: this is just the ground work and a subsequent PR will align all files (didn't want to disrupt other PRs already open).
* Small fixes

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,90 @@
+# This file is the top-most EditorConfig file
+root = true
+
+# All Files
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+##########################################
+# File Extension Settings
+##########################################
+
+# Visual Studio Solution Files
+[*.sln]
+indent_style = tab
+
+# Visual Studio XML Project Files
+[*.{csproj,vbproj,vcxproj.filters,proj,projitems,shproj}]
+indent_size = 2
+
+# XML Configuration Files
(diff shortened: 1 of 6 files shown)


eladiw (approved):
lgtm

AsafMah:
While we are looking at the style - we do we have the using inside the namespace?
It's less convenient to work with, and it forces the ugly global:: syntax

AsafMah:
The editorconfig file is pretty barebones, it's a good strategy to have a small one that is easier to maintain,
but another strategy is to generate a big one from the IDE and work from there.
I'm fine with both

.editorconfig
```diff
@@ -0,0 +1,90 @@
...
+# CA1848: Use the LoggerMessage delegates
+dotnet_diagnostic.CA1848.severity = suggestion
+
+[*.{cs,csx,cake}]
+# 'var' preferences
+csharp_style_var_for_built_in_types = true:warning
+csharp_style_var_when_type_is_apparent = true:warning
+csharp_style_var_elsewhere = true:warning
+# others
+# csharp_style_namespace_declarations = file_scoped:warning
```
AsafMah, line 90: Why is this commented out?

AsafMah (approved):
Looks good

.editorconfig
tamirkamara: It's something to uncomment after most of the PRs will be closed as this is going to change every single file.

PR #116, opened 2022-01-10
Normalize Match All

AsafMah:
Change Matchall to return "true" instead of null, so it would work with complex kusto expressions.


eladiw (approved):

eladiw:
@tomconte this fixes the sub issue in the filters pr, where if you put * in the filter text box you get an error

tomconte (approved):
Looks good, thanks!

PR #117, opened 2022-01-10
add stddev aggregation

Mimetis:
Creating the standard deviation aggregation.
The pseudo kusto code generated query is:

```
let _data = database(\"dev\").kibana_data_flights 
 | where (['timestamp'] >= todatetime(\"2007-01-11T10:14:21.7430000Z\") and ['timestamp'] <= todatetime(\"2022-01-11T10:14:21.7430000Z\"));

(_data | summarize ['1%extended_stats%2']=pack(
  'count', count(),
  'min', min(['DistanceKilometers']),
  'max', max(['DistanceKilometers']),
  'avg', avg(['DistanceKilometers']),
  'sum', sum(['DistanceKilometers']),
  'sum_of_squares', sum(pow(['DistanceKilometers'], 2)),
  'variance_population', variancep(['DistanceKilometers']),
  'variance_sampling', variance(['DistanceKilometers']),
  'std_deviation_population', stdevp(['DistanceKilometers']),
  'std_deviation_sampling', stdev(['DistanceKilometers'])) | as aggs);

(_data | count | as hitsTotal);

(_data | limit 0 | as hits)

```
We are calculating the upper bounds and lower bounds directly inside the code to get the bounds.
The formula used is coming from Elastic Search source code here:

https://github.com/elastic/elasticsearch/blob/0699c9351f1439e246d408fd6538deafde4087b6/server/src/main/java/org/elasticsearch/search/aggregations/metrics/InternalExtendedStats.java#L187-L194

``` java
     public double getStdDeviationBound(Bounds bound) {
        return switch (bound) {
            case UPPER, UPPER_POPULATION -> getAvg() + (getStdDeviationPopulation() * sigma);
            case UPPER_SAMPLING -> getAvg() + (getStdDeviationSampling() * sigma);
            case LOWER, LOWER_POPULATION -> getAvg() - (getStdDeviationPopulation() * sigma);
            case LOWER_SAMPLING -> getAvg() - (getStdDeviationSampling() * sigma);
        };
    }
```

The result in Kibana:

image


K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -138,6 +143,53 @@ public static void AddAggregates(this AggregateDictionary aggregateDictionary, s
...
+        /// <returns><see cref="ExtendedStatsAggregate"></returns>
+        private static ExtendedStatsAggregate GetExtendedStatsAggregate(string columnName, DataRow row, ILogger logger)
+        {
+            logger.LogTrace("Get extended stats for standard deviation aggregate for {}", columnName);
+
+            var extendedStatsAggregate = new ExtendedStatsAggregate();
+
+            var jObject = (JObject)row[columnName];
+
+            if (jObject == null)
```
AsafMah, line 161: This actually can never happen. if the value of `row[columnName]` is null, it will return a DBNull object, and fail at the cast in 159

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -138,6 +143,53 @@ public static void AddAggregates(this AggregateDictionary aggregateDictionary, s
...
+            {
+                return extendedStatsAggregate;
+            }
+
+            if (jObject.ContainsKey(AggregationsConstants.StandardDeviationBoundsUpper))
+            {
+                var upper = jObject[AggregationsConstants.StandardDeviationBoundsUpper].Value<double?>();
+                extendedStatsAggregate.StandardDeviationBounds.Upper = upper ?? (double.IsNaN(upper.Value) ? null : upper);
```
AsafMah, line 169: Why the NaN checks?

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -138,6 +143,53 @@ public static void AddAggregates(this AggregateDictionary aggregateDictionary, s
...
+            }
+
+            if (jObject.ContainsKey(AggregationsConstants.StandardDeviationBoundsLower))
```
AsafMah, line 172: Is there ever a legitimate reason for these keys not to exist? if not, then throw

K2Bridge/KustoDAL/KustoQueryExecutor.cs
```diff
@@ -57,13 +57,26 @@ internal class KustoQueryExecutor : IQueryExecutor
...
         public static KustoConnectionStringBuilder CreateKustoConnectionStringBuilder(IConnectionDetails connectionDetails)
         {
-            var conn = new KustoConnectionStringBuilder(
-                connectionDetails.ClusterUrl,
-                connectionDetails.DefaultDatabaseName)
-                .WithAadApplicationKeyAuthentication(
-                    connectionDetails.AadClientId,
-                    connectionDetails.AadClientSecret,
-                    connectionDetails.AadTenantId);
+            KustoConnectionStringBuilder conn;
```
AsafMah, line 60: I assume that's from the other branch that slipped in?

K2Bridge/Models/Response/Aggregations/ExtendedStats/ExtendedStatsAggregate.cs
```diff
@@ -0,0 +1,32 @@
...
+// See LICENSE file in the project root for full license information.
+
+namespace K2Bridge.Models.Response.Aggregations
+{
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// Describes extended stats response element for standard deviation.
+    /// </summary>
+    public class ExtendedStatsAggregate : IAggregate
```
AsafMah, line 12: Our class is very barren compared to the one from kibana:
```json
{
  ...

  "aggregations": {
    "grades_stats": {
      "count": 2,
      "min": 50.0,
      "max": 100.0,
      "avg": 75.0,
      "sum": 150.0,
      "sum_of_squares": 12500.0,
      "variance": 625.0,
      "variance_population": 625.0,
      "variance_sampling": 1250.0,
      "std_deviation": 25.0,
      "std_deviation_population": 25.0,
      "std_deviation_sampling": 35.35533905932738,
      "std_deviation_bounds": {
        "upper": 125.0,
        "lower": 25.0,
        "upper_population": 125.0,
        "lower_population": 25.0,
        "upper_sampling": 145.71067811865476,
        "lower_sampling": 4.289321881345245
      }
    }
  }
}
```
Do you know if we need the other fields?
Should we add them just for parity?

K2Bridge/Utils/AggregationsConstants.cs
```diff
@@ -11,5 +11,9 @@ public static class AggregationsConstants
         /// Only used for percentiles metrics at the moment
         /// </summary>
         public const char MetadataSeparator = '%';
+        public const string Average = "average";
```
AsafMah, line 14: Add docs

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
```diff
@@ -33,6 +34,29 @@ public void Visit(CardinalityAggregation cardinalityAggregation)
...
+            EnsureClause.StringIsNotNullOrEmpty(extendedStatsAggregation.Field, extendedStatsAggregation.Field, ExceptionMessage);
+
+            var sep = AggregationsConstants.MetadataSeparator;
+            var key = $"['{extendedStatsAggregation.Key}{sep}extended_stats']";
+            var average = $"{KustoQLOperators.Avg}({EncodeKustoField(extendedStatsAggregation)})";
+            var standard_deviation = $"{KustoQLOperators.StDev}({EncodeKustoField(extendedStatsAggregation)})";
+
+            var query = new StringBuilder();
+
+            query.Append($"{key}={KustoQLOperators.Pack}(");
```
AsafMah, line 50: Is there a good reason to use the array pack trick? why aren't they just normal columns?

K2Bridge.Tests.UnitTests/Visitors/AggregationVisitorTests.cs
```diff
@@ -26,7 +26,8 @@ public string AggregationVisit_WithEmptyAgg_ReturnsNoPrimary()
         [TestCase(ExpectedResult = "['A']=avg(['fieldA'])")]
         public string AggregationVisit_WithAvgAgg_ReturnsAvgPrimary()
         {
-            var aggregateClause = new AggregationContainer() {
+            var aggregateClause = new AggregationContainer()
```
AsafMah, line 29: This file just has formatting changes, please reset it and we'll have it in another PR

K2Bridge.Tests.UnitTests/JsonConverters/AggregationContainerConverterTests.cs
```diff
@@ -288,6 +297,24 @@ public class AggregationContainerConverterTests
             },
         };
 
+
+        private static readonly AggregationContainer ExpectedValidStandardDeviationAggregation = new AggregationContainer()
```
AsafMah, line 301: Add more tests, specifically visitor tests

AsafMah (changes requested):

K2Bridge/Factories/AggregateFactory.cs
Mimetis: Well, we have no control on the networking, on ADX and so on...
So far, theorically, this shoud not happened, but just in case, having this assumption could avoid some potential issues.

K2Bridge/KustoDAL/KustoQueryExecutor.cs
Mimetis: Good catch !
Indeed, I will remove it, sorry for that !

K2Bridge/Models/Response/Aggregations/ExtendedStats/ExtendedStatsAggregate.cs
Mimetis: We did not find any cases where the population and sampling values are used.

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
Mimetis: Because we need to have all the information in the same key (like `['1']` or `['2']` and so on )
All these values belongs to the standard deviation.
It's more convenient to have just one column when you deserialize everything

K2Bridge.Tests.UnitTests/Visitors/AggregationVisitorTests.cs
Mimetis: Ok

Mimetis: Fixed

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
```diff
@@ -33,6 +34,29 @@ public void Visit(CardinalityAggregation cardinalityAggregation)
...
+            query.Append($"'{AggregationsConstants.StandardDeviation}', {standard_deviation},");
+            query.Append($"'{AggregationsConstants.Average}', {average},");
+            query.Append($"'{AggregationsConstants.StandardDeviationBoundsUpper}', {average} + {standard_deviation} * {extendedStatsAggregation.Sigma},");
+            query.Append($"'{AggregationsConstants.StandardDeviationBoundsLower}', {average} - {standard_deviation} * {extendedStatsAggregation.Sigma}");
+            query.Append($")");
```
tamirkamara, line 55:
```suggestion
            query.Append(')');
```

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
```diff
@@ -33,6 +34,29 @@ public void Visit(CardinalityAggregation cardinalityAggregation)
...
         }
 
+        /// <inheritdoc/>
+        public void Visit(ExtendedStatsAggregation extendedStatsAggregation)
+        {
+            Ensure.IsNotNull(extendedStatsAggregation, nameof(extendedStatsAggregation));
+            EnsureClause.StringIsNotNullOrEmpty(extendedStatsAggregation.Field, extendedStatsAggregation.Field, ExceptionMessage);
+
+            var sep = AggregationsConstants.MetadataSeparator;
+            var key = $"['{extendedStatsAggregation.Key}{sep}extended_stats']";
```
tamirkamara, line 44: Please use EncodeKustoField here too. Something similar to:

```suggestion
            var key = EncodeKustoField(extendedStatsAggregation.Key + AggregationsConstants.MetadataSeparator + "extended_stats");
```

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
```diff
@@ -33,6 +34,29 @@ public void Visit(CardinalityAggregation cardinalityAggregation)
             cardinalityAggregation.KustoQL = $"{EncodeKustoField(cardinalityAggregation.Key)}={KustoQLOperators.DCount}({EncodeKustoField(cardinalityAggregation)})";
         }
 
+        /// <inheritdoc/>
+        public void Visit(ExtendedStatsAggregation extendedStatsAggregation)
+        {
+            Ensure.IsNotNull(extendedStatsAggregation, nameof(extendedStatsAggregation));
+            EnsureClause.StringIsNotNullOrEmpty(extendedStatsAggregation.Field, extendedStatsAggregation.Field, ExceptionMessage);
+
+            var sep = AggregationsConstants.MetadataSeparator;
```
tamirkamara, line 43: nit: redundant assignment.

tamirkamara (changes requested):

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
```diff
@@ -33,6 +34,29 @@ public void Visit(CardinalityAggregation cardinalityAggregation)
...
+            var key = $"['{extendedStatsAggregation.Key}{sep}extended_stats']";
+            var average = $"{KustoQLOperators.Avg}({EncodeKustoField(extendedStatsAggregation)})";
+            var standard_deviation = $"{KustoQLOperators.StDev}({EncodeKustoField(extendedStatsAggregation)})";
```
AsafMah, line 46: camel case

K2Bridge/Factories/AggregateFactory.cs
AsafMah: My point was to throw an error if these keys don't exist.
Silently discarding them is a bad option.

K2Bridge/Models/Response/Aggregations/ExtendedStats/ExtendedStatsAggregate.cs
AsafMah: Ok, but maybe leave a comment or something that we are aware that we don't have full parity to kibana

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
AsafMah: I see your point @eladiw @tomconte @dupuyjs @nebrass  WDYT?

K2Bridge/Factories/AggregateFactory.cs
AsafMah, line 169: Why the special nan handling? I tried giving it NaN and it doesn't return an error or anything. You can remove this check

K2Bridge/Models/Response/Aggregations/ExtendedStats/ExtendedStatsAggregate.cs
Mimetis: i'm working on it. I've found how to retrieve these values from ADX

K2Bridge/Factories/AggregateFactory.cs
Mimetis: Done

K2Bridge/Factories/AggregateFactory.cs
Mimetis: Done

K2Bridge/Models/Response/Aggregations/ExtendedStats/ExtendedStatsAggregate.cs
Mimetis: Done

K2Bridge/Utils/AggregationsConstants.cs
Mimetis: Done

K2Bridge.Tests.UnitTests/JsonConverters/AggregationContainerConverterTests.cs
Mimetis: Done

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
Mimetis: Done

K2Bridge/Factories/AggregateFactory.cs
Mimetis: If we want to have the exact same payload from elastic and from adx, we must handle Nan values.
I mean, when Elastic is returning null, ADX is returning NaN.
So far, handling the NaN value ensures we have the same payload

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
Mimetis: No, because EncodeKustoField will not return the expected value here

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
Mimetis: Done

K2Bridge/Models/Response/Aggregations/ExtendedStats/ExtendedStatsAggregate.cs
AsafMah: Great!

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
AsafMah: Why? is there a decimal point there?

K2Bridge/Factories/AggregateFactory.cs
AsafMah: Ok, understood.

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
AsafMah: Now it seems that all of them are snake_case instead of camel case:
var sum_of_squares = $"{KustoQLOperators.Sum}({KustoQLOperators.Pow}({EncodeKustoField(extendedStatsAggregation)}, 2))";
            var variance_population = $"{KustoQLOperators.VariancePopulation}({EncodeKustoField(extendedStatsAggregation)})";
            var variance_sampling = $"{KustoQLOperators.Variance}({EncodeKustoField(extendedStatsAggregation)})";
            var standard_deviation_population = $"{KustoQLOperators.StDevPopulation}({EncodeKustoField(extendedStatsAggregation)})";
            var standard_deviation_sampling = $"{KustoQLOperators.StDev}({EncodeKustoField(extendedStatsAggregation)})";

please fix

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
dupuyjs: No objection for this trick. If you consider percentiles scenario, percentiles_array returns all values in the same column. Here, this is the same concept. The only potential issue will be to lost the associated data type (as we now have a json). But here all values are double and pack everything in one only column really help during response parsing.

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
Mimetis: Ah ! Indeed, we had a point previously, but we've changed it.
Good catch ! thx :)
Done

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
Mimetis: Done

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -167,57 +167,57 @@ private static ExtendedStatsAggregate GetExtendedStatsAggregate(string columnNam
             var jObject = (JObject)row[columnName];
 
             var count = jObject[AggregationsConstants.Count].Value<long?>();
-            extendedStatsAggregate.Count = count ?? (double.IsNaN(count.Value) ? null : count);
+            extendedStatsAggregate.Count = count.HasValue ? double.IsNaN(count.Value) ? null : count : null;
```
AsafMah, line 170: Why this change? didn't `??` worked fine?

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -167,57 +167,57 @@ private static ExtendedStatsAggregate GetExtendedStatsAggregate(string columnNam
             var jObject = (JObject)row[columnName];
 
             var count = jObject[AggregationsConstants.Count].Value<long?>();
```
AsafMah, line 169: Since you are nullchecking anyway, you can have:
`jObject[AggregationsConstants.Count]?.Value<long?>();` and avoid a potential NRE (and for the others too)

K2Bridge/Factories/AggregateFactory.cs
Mimetis: Nop; it's raising an error here.
I've made some more tests, and the only way to do it working is to use the `HasValue` property

Mimetis: image

K2Bridge/Factories/AggregateFactory.cs
Mimetis: good catch !

AsafMah: I think that because of that you can go back to ?? before?

Mimetis: No, Look my screenshot on the previous comment.
Even using a `double?` count will not allow us to use `??`

K2Bridge.Tests.End2End/TestElasticClient.cs
```diff
@@ -90,6 +90,40 @@ public static void DeleteValuesToComparePercentiles(JToken parent)
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Make a Math.Round operation on all float values in a json payload.
+        /// </summary>
+        public static void RoundFloats(JToken parent, string jsonPath, int? digits = null)
+        {
+            if (digits.HasValue)
```
eladiw, line 115: if (!digits.HasValue){
  return;
}

Mimetis: In my projects I like to use return as soon as possible.
But in most of the projects I've been working so far, developers don't like the return key word (for some reasons)

But if it's something you are comfortable with, I'm in 100% !

K2Bridge.Tests.End2End/flights/MSearch_Viz_ExtendedStats.json
```diff
@@ -0,0 +1,50 @@
+{
+    "aggs": {
```
eladiw, line 2: this scenario is when there is no bucket?

K2Bridge.Tests.UnitTests/JsonConverters/AggregationContainerConverterTests.cs
```diff
@@ -377,6 +396,41 @@ public class AggregationContainerConverterTests
...
+            PrimaryAggregation = null,
+            SubAggregations = new AggregationDictionary
+            {
+                ["2"] = new AggregationContainer
+                {
+                    PrimaryAggregation = new ExtendedStatsAggregation
+                    {
+                        Field = "metric",
+                        Key = "2",
+                        Sigma = 3,
```
eladiw, line 427: btw, what is sigma?

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -138,6 +143,88 @@ public static void AddAggregates(this AggregateDictionary aggregateDictionary, s
...
+            logger.LogTrace("Get extended stats for standard deviation aggregate for {}", columnName);
+
+            var extendedStatsAggregate = new ExtendedStatsAggregate();
+
+            if (row[columnName] == DBNull.Value)
+            {
+                return extendedStatsAggregate;
+            }
+
+            var sigma = Convert.ToInt32(columnMetadata[2]);
```
eladiw, line 165: can we move '2' to an enum?

eladiw: same as columnName

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -138,6 +143,88 @@ public static void AddAggregates(this AggregateDictionary aggregateDictionary, s
...
+
+            extendedStatsAggregate.StandardDeviationBounds.Upper = extendedStatsAggregate.StandardDeviationBounds.UpperPopulation;
+
+            if (extendedStatsAggregate.StandardDeviationSampling.HasValue && extendedStatsAggregate.Average.HasValue)
+            {
+                extendedStatsAggregate.StandardDeviationBounds.UpperSampling = extendedStatsAggregate.Average + (extendedStatsAggregate.StandardDeviationSampling.Value * sigma);
+            }
+
+            return extendedStatsAggregate;
+        }
```
eladiw, line 226: looks good (this file)

K2Bridge.Tests.UnitTests/JsonConverters/AggregationContainerConverterTests.cs
Mimetis: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-extendedstats-aggregation.html#_standard_deviation_bounds

By default, the `extended_stats` metric will return an object called `std_deviation_bounds`, which provides an interval of plus/minus two standard deviations from the mean. 
This can be a useful way to visualize variance of your data. If you want a different boundary, for example three standard deviations, you can set `sigma` in the request

K2Bridge/Factories/AggregateFactory.cs
Mimetis: We have this index used in several aggregations so far.
I guess it could be a good idea to move it to an enum, but it should be part of a global PR to get something homogenous for all aggregations
your thought ?

K2Bridge.Tests.End2End/flights/MSearch_Viz_ExtendedStats.json
Mimetis: I think so

K2Bridge/Factories/AggregateFactory.cs
Mimetis: Should I do something here ?

K2Bridge/Models/Request/Aggregations/Metric/ExtendedStats/ExtendedStatsAggregation.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <summary>
+    /// A single-value metrics aggregation that provides an interval of plus/minus two standard deviations from the mean
+    /// </summary>
+    internal class ExtendedStatsAggregation : MetricAggregation
+    {
+        /// <summary>
+        /// Gets or sets Sigma field
+        /// </summary>
+        [JsonProperty("sigma")]
+        public double Sigma { get; set; } = 2;
```
eladiw, line 19: what doess 2 means? (and what does 3?)

K2Bridge/Factories/AggregateFactory.cs
eladiw: no

K2Bridge/Models/Request/Aggregations/Metric/ExtendedStats/ExtendedStatsAggregation.cs
eladiw: Its the default value...

eladiw (approved):
LGTM!

K2Bridge/Visitors/Aggregations/MetricAggregationVisitor.cs
AsafMah: Ok, closing

K2Bridge/Factories/AggregateFactory.cs
AsafMah: Ok I see, I'll mark this as resolved and continue in the other one

K2Bridge/Factories/AggregateFactory.cs
AsafMah: Ok so the problem is the IsNaN.
So first of all you don't have to check it for the long.

Second of all, I'm not a fan of the double ternarys personally.

What do you think about the alternatives:
1.
```c#
extendedStatsAggregate.Min = (min.HasValue && !double.IsNaN(min.Value)) ? min.Value : null;
```
2.
```c#
var min = jObject[AggregationsConstants.Min]?.Value<double?>() ?? double.NaN;
extendedStatsAggregate.Min = double.IsNaN(min) ? null : min;
```
3.
```c#
extendedStatsAggregate.Min = (min is double d && !double.IsNaN(d)) ? d : null;
```
4.
```c#
extendedStatsAggregate.Min = (min is {} d && !double.IsNaN(d)) ? d : null;
```

Mimetis: Your first solution is the best one in my opinion.
Good catch !
As you I was not a big fan of the double ternarys check
Damn, I can't believe I did not found that solution ! :)

K2Bridge.Tests.End2End/TestElasticClient.cs
```diff
@@ -182,6 +218,9 @@ public async Task<JToken> MSearch(string indexName, string jsonQueryFile, bool v
             // Normalize aggregate value (double) with fixed number of decimal
             NormalizeAggregateValue(result, "responses[*].aggregations..value");
 
+            // make a Math.Round on all floats values
+            RoundFloats(result, "responses[*].aggregations", roundingFloats);
+
```
dupuyjs, line 223: Ok. Nice, it means NormalizeAggregateValue is probably not needed any longer. You can remove it.

dupuyjs (approved):
Approved, just one comment about NormalizeAggregateValue not needed any longer.

AsafMah (approved):
Looks good!
Liked the RoundFloats solution

K2Bridge/Factories/AggregateFactory.cs
eladiw: ok, please open an issue

K2Bridge.Tests.End2End/TestElasticClient.cs
dupuyjs: Forget. We will see that in another PR.

tamirkamara (approved):
Visitor looks good (didn't check other parts)

PR #119, opened 2022-01-10
Upgrade to ES 7.16.2

eladiw:

* Upgraded to ES 7.16.2
* Added new field caps field [metadata_field](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-fields.html)
* fixed tests


tomconte (approved):

charts/k2bridge/values.yaml
```diff
@@ -39,8 +39,8 @@ elasticsearch:
   clusterName: k2bridgees
   masterService:  k2bridgees-master
 
-  image: docker.elastic.co/elasticsearch/elasticsearch-oss
-  imageTag: 7.10.2
+  image: docker.elastic.co/elasticsearch/elasticsearch
+  imageTag: 7.16.2
```
tamirkamara, line 43: Is it perhaps the right time to also update the elastic chart version we use?

AsafMah: what's elastic chart?

AsafMah (changes requested):
I understand the version changes, but why add metadata_field if it's always false?

eladiw:
> I understand the version changes, but why add metadata_field if it's always false?

to match the results ES returns.
maybe Kibana use it, maybe not.
In anycase the e2e failed, since now our reference ES instance returns it and K2 didn't.

charts/k2bridge/values.yaml
eladiw: updated.

@AsafMah , its the chart the defines how to deploy elastic. we use it as a dependency
https://github.com/elastic/helm-charts/tree/main/elasticsearch

AsafMah (approved):

PR #122, opened 2022-01-11
Updated devcontainer

tamirkamara:

* Move to dotnet 6 base image (still focal/ubuntu20)


eladiw (approved):
tested. lgtm

AsafMah:
Looks good, is there a reason to replace the run commands with &&?
I'm not an expert on docker, but I assume it would provide better logs/errors if each command was separate, no?

eladiw:
> Looks good, is there a reason to replace the run commands with &&? I'm not an expert on docker, but I assume it would provide better logs/errors if each command was separate, no?

Less RUN commands results in less layers which is better practice.

AsafMah (approved):

PR #123, opened 2022-01-12
Delegate full KustoQL expression to aggs visitors

tomconte:

Prior to working on the fix for Filters, I would like to delegate the full KustoQL expression creation to the bucket aggregation visitors, i.e. allow them to generate a full KustoQL statement and not only the part that goes in "(... | as aggs)".

This way the visitors could insert multiple statements in their KustoQL.

I also moved the DateRange source files to their proper location.


AsafMah:
I see the utility, but I don't like the code duplication.
What do you think of having a method "WrapWithFullStatement"(bad name, but you get the point) in the base class that all the aggregators will call, and it wall add the ( ... | as aggs) thing.

If not, can you maybe split it into different "append" class? because it's very easy to miss and I think it's error prone.
Instead of 
```c#
            queryStringBuilder.Append($"({KustoTableNames.Data} | {KustoQLOperators.Extend} {EncodeKustoField(filtersAggregation.Key)} = {KustoQLOperators.PackArray}(");
```
```c#
            queryStringBuilder.Append('(');
            queryStringBuilder.Append($"{KustoTableNames.Data} | {KustoQLOperators.Extend} {EncodeKustoField(filtersAggregation.Key)} = {KustoQLOperators.PackArray}(");
```
And in the closing too

AsafMah (approved):
LGTM

eladiw (approved):
LGTM

Issue #120, opened 2022-01-10
Filter aggregation fails when there are no matches


eladiw:
This bug is about the visualization part.
Reproduction:

1. create a new visualization graph
2. select filter aggregation (x axis)
3. select multiple values, that do no exists, so the results will be blank
4. it will fail since the generate KustoQL is not generating buckets when nothing is returned.

see complete thread here:
https://github.com/microsoft/K2Bridge/pull/114#issuecomment-1008979540

PR #125, opened 2022-01-13
Handle missing buckets for Filters aggregation

tomconte:

* Fix for issue #120 

This fix does not impact the query structure. It uses the same trick used in other aggregations to encode all the expected bucket names in the aggregation column. When parsing the response, any missing buckets are added with a doc count of 0.

Since the bucket names are Lucene search patterns, they are encoded use Base64 to avoid any illegal characters in the column name. This limits the total combined size of the filters in an aggregation to about ~770 characters (taking in consideration the max entity length in Kusto of 1024 characters, an UTF-8 encoding of one byte for most common characters, and the 33% size increase due to Base64).


AsafMah:
Very creative!

One question though - why store it in the column's name, instead of the columns data?

like instead of `project-rename ['2%RGVzdFdlYXRoZXI6Kg--%RGVzdFdlYXRoZXI6cio-%RGVzdFdlYXRoZXI6c2hsb21v']=['_tmp']`
do `| extend filters "<the filters in any format you want>"`?
If you worry about duplication, just do:
`| extend filters = iff(row_number() == 1, "<the filters in any format you want>", "")`

And then they'll only be on the first row.

tomconte:
> Very creative!

Thanks! Definitely abusing this column name trick 😆

> One question though - why store it in the column's name, instead of the columns data?
> 
> like instead of `project-rename ['2%RGVzdFdlYXRoZXI6Kg--%RGVzdFdlYXRoZXI6cio-%RGVzdFdlYXRoZXI6c2hsb21v']=['_tmp']` do `| extend filters "<the filters in any format you want>"`? If you worry about duplication, just do: `| extend filters = iff(row_number() == 1, "<the filters in any format you want>", "")`
> 
> And then they'll only be on the first row.

I did not want to duplicate that data on every row indeed but did not think of the row_number() trick! I will give it a try, it should simplify the code much.

tomconte:
> like instead of `project-rename ['2%RGVzdFdlYXRoZXI6Kg--%RGVzdFdlYXRoZXI6cio-%RGVzdFdlYXRoZXI6c2hsb21v']=['_tmp']` do `| extend filters "<the filters in any format you want>"`? If you worry about duplication, just do: `| extend filters = iff(row_number() == 1, "<the filters in any format you want>", "")`

I played with this and ran into a problem: what if there are no matches to any filter? Then there is not even one row where we can store the full filter list.

tomconte:
I am exploring another idea that could be fairly generic:

```
(_data
 | extend ['2'] = pack_array('a','c'), ['_filter_value'] = pack_array((* has "a"),(* has "c"))
 | fork
    aggs = (mv-expand ['2'] to typeof(string), ['_filter_value']
            | where ['_filter_value'] == true
            | summarize ['3']=avg(['AvgTicketPrice']), count() by ['2']
            | order by ['2'] asc )
    buckets = (take 1 | project-keep ['2'])
);
```

tomconte:
No with this option I would revert the changes that use the column name: since all the possible bucket names (filters in this case) from the original `pack_array()` are stored in the "buckets" table, in JSON format, I have all I need to assemble a full list of response buckets.

That said, I am still not sure this new query is compatible with the changes in @dupuyjs PR, so in the short term, I am afraid the column trick is the only workaround I have.

tamirkamara:
I remember something about **fork** not being recommended or that the newer way of batching (semicolon separated statements) is preferred. Not 100% sure though... better check that out before investing too much in this direction?

K2Bridge/Factories/BucketFactory.cs
```diff
@@ -129,7 +129,9 @@ public static FiltersBucket CreateFiltersBucket(string primaryKey, DataRow row,
         {
             Ensure.IsNotNull(row, nameof(row));
 
-            var filter = Convert.ToString(row[primaryKey]);
+            var encodedKey = row.Table.Columns[0].ColumnName;
```
eladiw, line 132: this change is not 100% clear to me

eladiw:
Looks good. Waiting to see the new version based on @AsafMah  comments

dupuyjs:
> I am exploring another idea that could be fairly generic:
> 
> ```
> (_data
>  | extend ['2'] = pack_array('a','c'), ['_filter_value'] = pack_array((* has "a"),(* has "c"))
>  | fork
>     aggs = (mv-expand ['2'] to typeof(string), ['_filter_value']
>             | where ['_filter_value'] == true
>             | summarize ['3']=avg(['AvgTicketPrice']), count() by ['2']
>             | order by ['2'] asc )
>     buckets = (take 1 | project-keep ['2'])
> );
> ```

In my opinion, the problem is not query we are building, but the way we are parsing the response. If there is no valid answer, we should be able to build the response by taking the information (names of buckets) in the DSL object. 

At the moment, until we cannot do this, I suggest to retrieve names of the buckets as part of the column name (as we are doing for percentiles).

dupuyjs (approved):
Good for me until we keep changes in the column names. We know we will be able to do remove this 'hack' the day we can grab data from the DSL object (or at least the aggregations part form the DSL object) .

eladiw:
@tomconte are you working on a new version or as @dupuyjs suggested, keep this version with the column name trick?

tomconte:
> @tomconte are you working on a new version or as @dupuyjs suggested, keep this version with the column name trick?

I think we will stay with this version. I will address comments today.

K2Bridge/Factories/BucketFactory.cs
tomconte: I have added some comments to clarify

AsafMah (approved):
Looks good! please add a test for an empty filter result if possible

PR #127, opened 2022-01-16
Add SummarizableMetricAggregation base class

dupuyjs:

Add SummarizableMetricAggregation base class. This class defines if a metric aggregation can be used with summarize operator. Applied on Average, Cardinality, ExtendStats, Max, Min, Percentile and Sum.


AsafMah:
So this interface's only goal is to mark which metrics can be summerizable, which right now are everyone except top hits?

eladiw (approved):
LGTM I would add a section in the docs to document all of this knowledge

dupuyjs:
> So this interface's only goal is to mark which metrics can be summerizable, which right now are everyone except top hits?

Yes, this is the exact and only purpose, ISummarizable interface identifies metrics where summarize can be applied. TopHits will be the only one with another interface (IPartitionable) at the moment.

AsafMah:
I wrote it in the other one too, I'm not a big fan of marker interfaces, and here especially because you only mark MetricAggregations.
Is there a reason we couldn't have it as part of the hierarchy (SummarizableMetricAggragegation inherits from MetricAggregation)?

dupuyjs:
> I wrote it in the other one too, I'm not a big fan of marker interfaces, and here especially because you only mark MetricAggregations. Is there a reason we couldn't have it as part of the hierarchy (SummarizableMetricAggragegation inherits from MetricAggregation)?

No objection. Let me check.

K2Bridge/Visitors/Aggregations/AggregationContainerVisitor.cs
```diff
@@ -36,13 +37,17 @@ public string BuildSummarizableMetricsQuery(AggregationDictionary aggregationDic
         {
             var query = new StringBuilder();
 
-            // Collect all metrics
+            // Collect all ISummarizable metrics
```
AsafMah, line 40: It's no longer ISummarizable.

K2Bridge/Visitors/Aggregations/AggregationContainerVisitor.cs
```diff
@@ -36,13 +37,17 @@ public string BuildSummarizableMetricsQuery(AggregationDictionary aggregationDic
...
             // ['2']=max(AvgTicketPrice), ['3']=avg(DistanceKilometers)
             var summarizableMetrics = new List<string>();
-            foreach (var (_, aggregation) in aggregationDictionary)
+            foreach (var (_, aggregationContainer) in aggregationDictionary)
```
AsafMah, line 43: Get the revised method from 3/3

K2Bridge/Visitors/Aggregations/AggregationContainerVisitor.cs
dupuyjs: Fixed.

K2Bridge/Visitors/Aggregations/AggregationContainerVisitor.cs
dupuyjs: Done.

AsafMah (approved):

PR #134, opened 2022-01-19
Fix missing buckets issue for Range and Filters

tomconte:
This is a new fix to solve the missing buckets issue for both Filters and Ranges. The problem happens when one of the expressions (filter or range) does not match any rows. In this case the bucket is completely missing from the Kusto response, and missing in the generated JSON, which triggers errors in Kibana.

This fix uses the new query structure in Visitors to add a statement to Kusto QL queries for Filters and Ranges. This statement creates a "metadata" table that contains all expected bucket name values.

Example for filters:

```
datatable(key:string, value:string) ['key','foo:bar','key','bar:baz'] | as metadata;
```

Example for ranges:

```
datatable(key:string, value:string) ['key','0%800','key','800%2000'] | as metadata;
```

This new table is used in the response parser, to identify any missing buckets in the response, and generate empty ones.

This is a more generic fix that could be used to fix similar problems in other aggregations in the future, including metrics, and avoid the column naming trick.

The previous fix for Filters based on the column name trick (#125) has been reversed in this PR.


K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -53,21 +54,42 @@ public static BucketAggregate GetDateHistogramAggregate(string key, DataTable da
...
                 var bucket = BucketFactory.CreateRangeBucket(key, row, logger);
                 if (bucket != null)
                 {
                     rangeAggregate.Buckets.Add(bucket);
+                    outputBuckets.Add(Convert.ToString(row[key]));
                 }
             }
 
+            // Get expected bucket names from metadata row
+            var metadataJson = (string)metadataTable.Rows[0][key];
```
AsafMah, line 78: This code is very similar to range, can we maybe extract it to a common method?

K2Bridge/Visitors/Aggregations/AggregationDictionaryVisitor.cs
```diff
@@ -70,6 +71,47 @@ public string BuildBucketAggregationQuery(BucketAggregation bucketAggregation, B
...
+            return query.ToString();
+        }
+
+        /// <summary>
+        /// Given a metadata dictionary, generates a Kusto QL statement creating the metadata table:
+        /// datatable(['2']:string) [dynamic(['val1','val2', 'val3'])] | as metadata;
+        /// </summary>
+        /// <param name="metadata">A metadata dictionary.</param>
+        /// <returns>A string containing the Kusto QL statement.</returns>
+        public string BuildMetadataQuery(Dictionary<string, string> metadata)
```
AsafMah, line 89: Thought - instead of building one dynamic cell that you then parse as json,
why not have a row in the table for each value? seems less "hacky"

AsafMah, line 89: For now metadata only has one column.
Do you expect to sometime need multiple columns?

AsafMah (commented):
Really inventive and elegeant solution!

Some points for discussion

AsafMah:
@tomconte I think I found a bug (not sure if it's introduced in this PR).
Ranges that have negative numbers in them return errors for me, I assume it has to do with the parsing (-5--3), please take a look.

tomconte:
Good catch!! It has probably been there all this time 😅 I think I will switch to the usual separator '%'.

K2Bridge/Visitors/Aggregations/AggregationDictionaryVisitor.cs
```diff
@@ -70,6 +71,47 @@ public string BuildBucketAggregationQuery(BucketAggregation bucketAggregation, B
             query.Append($"{KustoQLOperators.CommandSeparator} {KustoQLOperators.Summarize} {bucketAggregation.SubAggregationsKustoQL}");
             query.Append($"{definition.BucketExpression};");
 
+            // Create a table named "metadata", with one column per aggregation and dynamic value.
```
eladiw, line 74: I am missing a bit more elaboration here. other wise looks good

K2Bridge/Visitors/Aggregations/AggregationDictionaryVisitor.cs
dupuyjs: Yes, if we plan to use the same logic for Metrics. It means we will have one additional column per metric with metadata.

K2Bridge/KustoDAL/KustoResponseParser.cs
```diff
@@ -168,6 +168,13 @@ private void ReportQueryExecutionMetrics(KustoResponseDataSet kustoResponseDataS
                 var (key, aggregationType) = query.PrimaryAggregation;
                 var tableData = kustoResponse[KustoTableNames.Aggregation].TableData;
 
+                // Not all aggregations return a metadata table
+                DataTable metadataTableData = null;
```
dupuyjs, line 172: I suppose you can replace all this block by:
var metadataTableData = kustoResponse[KustoTableNames.Metadata]?.TableData;

K2Bridge/Visitors/Aggregations/AggregationDictionaryVisitor.cs
tomconte: Right, I was trying to avoid "querying" the table, i.e. find all the rows where the key matches and concatenate the values. But maybe that is not that complicated. I will give it a try.

K2Bridge/KustoDAL/KustoResponseParser.cs
tomconte: Fixed

K2Bridge/Visitors/Aggregations/AggregationDictionaryVisitor.cs
tomconte: Can you check the comment for BuildMetadataQuery()? It should clarify.

K2Bridge/Visitors/Aggregations/AggregationDictionaryVisitor.cs
tomconte: I have changed the schema to a more traditional key/value scheme. LMKWYT

K2Bridge/Factories/AggregateFactory.cs
tomconte: Done

K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -56,21 +59,29 @@ public static BucketAggregate GetDateHistogramAggregate(string key, DataTable da
...
                 var bucket = BucketFactory.CreateRangeBucket(key, row, logger);
                 if (bucket != null)
                 {
                     rangeAggregate.Buckets.Add(bucket);
+                    outputBuckets.Add(Convert.ToString(row[key]));
                 }
             }
 
+            // Add missing buckets
+            AddMissingBuckets(key, dataTable, metadataTable, outputBuckets, rangeAggregate, BucketFactory.CreateRangeBucket, logger);
```
dupuyjs, line 83: Personally, I prefer to keep directly the foreach done by AddMissingBuckets inside the Aggregate and avoid the delegate usage. OK, it will add a couple of lines but will be more readable.

dupuyjs (approved):
LGTM. Just added a comment on the delegate.

K2Bridge/Factories/AggregateFactory.cs
tomconte: Yeah, it's a bit heavy 😬 We might have to use it also for Date Range at some point. I'm happy both ways :) @AsafMah

K2Bridge/Visitors/KustoQLOperators.cs
```diff
@@ -73,6 +73,8 @@ public static class KustoQLOperators
         public const string CommandSeparator = "\n|";
         public const string NewLine = "\n";
         public const string True = "true";
+        public const string Datatable = "datatable";
+        public const string Dynamic = "dynamic";
```
AsafMah, line 77: No longer used

K2Bridge/Visitors/Aggregations/AggregationDictionaryVisitor.cs
AsafMah: Which metrics are expected to have metadata? will they still be on the same aggregation key?

AsafMah: It looks good to me. This is more performant (but probably neglegable since they will probably not be more than a handful of filters/ranges), but I think it's more readable/understandable. Do you agree or do you still prefer the dynamic way?

K2Bridge/Factories/AggregateFactory.cs
AsafMah: I'm really torn on this too, 7 parameters is a lot.
For now it seems that we don't need it in date range, right? (they work fine with empty buckets).
So I think maybe go back to duplicating the lines, and if we'll need it 3 times we might want to do a more serious refactor (maybe put it into data types instead of static functions).

But your choice really.

K2Bridge/Visitors/KustoQLOperators.cs
tomconte: Fixed

tomconte:
Metrics that need to use the metadata table would insert their own aggregation key. i.e. if the bucket aggregation key is '2', the percentiles metrics key could be '3'. Currently the metrics that use the column name trick are Percentiles and Extended Stats if I'm not mistaken.

I am happy with the new table approach, the dynamic column was indeed a bit obscure!

K2Bridge/Visitors/Aggregations/AggregationDictionaryVisitor.cs
tomconte: Metrics that need to use the metadata table would insert their own aggregation key. i.e. if the bucket aggregation key is '2', the percentiles metrics key could be '3'. Currently the metrics that use the column name trick are Percentiles and Extended Stats if I'm not mistaken.

tomconte: I am happy with the new table approach, the dynamic column was indeed a bit obscure!

AsafMah: But for now we don't plan to change them to this, right? ( I guess we don't have much time?)

K2Bridge/Factories/AggregateFactory.cs
tomconte: OK, I moved back to duplicating the code, and we can decide in the future to refactor.

AsafMah (approved):
Looks great!

PR #148, opened 2022-01-23
Update stylecop version and minor fixes

tamirkamara:

* Stylecop version to support newer versions of c#
* Logging params in Pascel Case
* Using var rather than explicit types
* More specific exceptions
* Remove unnecessary using statements
* Fix issues with comments
* Make methods as static if possible


AsafMah:
I think this should wait until the top-hits PR is merged

tamirkamara:
Just seen there're many files there too... we're going to wait.

eladiw (approved):

PR #149, opened 2022-01-23
Update Stylecop

tamirkamara:

* Update stylecop to support latest version of c#
* Suppress all current issues which will be handled separately


AsafMah (approved):
Looks good.

eladiw (approved):

PR #153, opened 2022-01-24
Create codeql-analysis.yml

eladiw:

* Adding security code scanning


AsafMah:
Does it only do these two checks?

eladiw:
> Does it only do these two checks?

yes

AsafMah (approved):

tomconte (approved):

PR #156, opened 2022-01-25
Fixed using quotes in queries 

AsafMah:
# Query strings
Lucene query strings (which means the search bar in discovery and the filters in filter aggregation) now have the same escaping logic as kibana.

This solves the discussion in #50 

This means that for these strings, it will search:
`test` => `test`
`"test"` => `test`
`foo:"test"` => `foo:test`
`\"test\"` => `"test"`
`foo:\"test\"` => `foo:"test"`
`"test` => error
`\` => error
`\\` => `\`

Filter bucket names will also be escaped, to match kibana's behavior.

# Match clause

Match clause will search literally for the value, like in kibana:
`test` => `test`
`"test"` => `"test"`
`\` => `\`

This solves #26


eladiw (approved):
nice job

tomconte (approved):

PR #158, opened 2022-01-25
Add top hits aggregation documentation

dupuyjs:

* Add top hits aggregation documentation


eladiw (approved):

AsafMah (commented):
I know the description is copy-pasted from kibana, but I think it's a bad description that doesn't really help you to understand what it is.

I think it would be nice to have a line or two on what it does and maybe a photo

AsafMah (approved):
Looks good!

Just before we merge, let's make sure it's for the correct branch - if dev-vnext is still relevant or main

PR #172, opened 2022-01-26
Remove suppressions and fix code analyis issues

tamirkamara:

* Logging params should be is PascalCase
* Mark static when possible


eladiw (approved):

AsafMah (approved):
Looks good!

PR #173, opened 2022-01-26
Remove suppressions and fix code analysis issues (tests projects)

tamirkamara:

* Various code analysis fixes on the tests projects


K2Bridge/Factories/AggregateFactory.cs
```diff
@@ -458,7 +458,7 @@ private static TopHitsAggregate GetTopHitsAggregate(string columnName, string[]
             topHitsAggregate.Hits.SetTotal(0);
 
             var rowValue = row[columnName];
-            if (rowValue.GetType() != typeof(System.DBNull))
+            if (rowValue.GetType() != typeof(DBNull))
```
AsafMah, line 461: If we are touching this line anyway (and in the next one too)
```suggestion
            if (rowValue is not DBNull)
```

.editorconfig
```diff
@@ -67,6 +67,7 @@ dotnet_analyzer_diagnostic.category-Style.severity = warning
 dotnet_diagnostic.IDE0005.severity = warning    # Using directive is unnecessary.
 dotnet_diagnostic.CA1848.severity = none        # Do not encourage LoggerMessage delegates in every instance
 dotnet_diagnostic.CA2254.severity = suggestion  # Logging template should not vary between calls
+dotnet_diagnostic.IDE0058.severity = suggestion
```
AsafMah, line 70: Do we have tons of expressions that we ignore the result for?

K2Bridge.Tests.End2End/PrometheusTest.cs
```diff
@@ -60,23 +60,25 @@ public async Task PrometheusTelemetry_WhenQueryParsed_ThenExposeQueryMetrics()
                 responseData.Contains(
                 $"# HELP {metrics.AdxQueryDurationMetric.Name} {metrics.AdxQueryDurationMetric.Help}\n"
                 + $"# TYPE {metrics.AdxQueryDurationMetric.Name} histogram\n",
-                Ordinal), responseData);
+                Ordinal),
```
AsafMah, line 63: This indentation feels wrong, why aren't the inner rows indented more?

.editorconfig
tamirkamara: yes, probably hundreds 😞  
All the string builders but others as well...

eladiw (approved):

K2Bridge.Tests.End2End/PrometheusTest.cs
tamirkamara: [SA1137](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1137.md) insists that line 63 should be the same as 61. 62 can be be pushed by a tab.

AsafMah: When I format it, it turns into this (and that's how I think it should be):
```c#
            Assert.True(
                responseData.Contains(
                    "# HELP exceptions Exceptions logged\n"
                    + "# TYPE exceptions counter\n"
                    + "exceptions ",
                    Ordinal)
                );
```

I think we can control it with https://www.jetbrains.com/help/resharper/EditorConfig_CSHARP_CSharpIndentStylePageSchema.html#Parenthesis

`csharp_indent_method_decl_pars` and `csharp_indent_invocation_pars`

.editorconfig
AsafMah: Ok, I guess it's ok for now.
Add a comment though?

AsafMah (approved):
looks good (assuming we leave the tabs to another PR)

PR #177, opened 2022-01-28
Add percentile error estimation to documentation

nebrass:
Add percentile error estimation to documentation


eladiw (approved):
🚀

AsafMah (approved):
This is from the kusto doc -https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/percentiles-aggfunction

Probably good to have

PR #186, opened 2022-05-18
Fix sort on dynamic columns

fonsecamar:
Adding wrapDynamic=true to fix sorting on dynamic columns.


felipemoz:
love it!

AsafMah:
Pipeline has been fixed, but resetting it here doesn't work, so I'm merging manually.

